PR #2, opened 2015-10-07
Add wtime and wtick.

hestela:
These functions are used for timing performance.
Simple time usage example added to time barrier.

### DIFF
--- a/examples/time.rs
+++ b/examples/time.rs
@@ -1,0 +1,15 @@
+extern crate mpi;
+
+use mpi::traits::*;
+
+fn main() {
+    let universe = mpi::initialize().unwrap();
+    let world = universe.world();
+
+    let t_start = universe.get_time();
+    world.barrier();
+    let t_end = universe.get_time();
+
+    println!("barrier took: {} s", t_end - t_start);
+    println!("the clock has a resoltion of {} seconds", universe.get_time_res());
+}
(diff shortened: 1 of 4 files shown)


hestela:
This may not be the best place for these functions.
Maybe in a place for mpi utility functions like all of the get_* functions.

bsteinb:
Hi and thank you for the contribution. Just a few remarks:

- I would prefer for these functions to be made methods of the `Universe` object like `get_processor_name()`. `Universe` acts as a guard (although not a perfect one) to ascertain initialisation before using MPI functions. `get_version()` and `get_library_version()` are free-standing functions, because the standard permits calling them before `MPI_Init()`.
- This passes CI tests on MPICH and OpenMPI as it is. However, the MPI 3.1 standard allows for `MPI_Wtime()` and `MPI_Wtick()` to be implemented as C preprocessor macros which could cause trouble later on. In the long term I would like to capture these "functions" in the C shim (`rsmpi.{c|h}`) as `RSMPI_Wtime()`/`RSMPI_Wtick()` and use those from the Rust side.
- I would prefer to use `c_double` as the return type. Although it is just a `typedef` for `f64`, I think it is more "honest" here, if that makes any sense.

bsteinb:
> In the long term I would like to capture these "functions" in the C shim (`rsmpi.{c|h}`) as `RSMPI_Wtime()`/`RSMPI_Wtick()` and use those from the Rust side.

... meaning I would also merge this part of the PR as it is.

hestela:
should new c/h files be created for the purpose of functions that could become macros?
rsmpi.c seems to only hold consts.

bsteinb:
I think they can go in the existing files for now.

PR #3, opened 2015-12-20
How to run the README example.

thyrgle:
Maybe I am doing this incorrectly, but the example will not run with:

```bash
cargo build
cargo run
```

Instead I did:

```bash
cd target/debug
mpirun -np 2 ./<project_name>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,13 @@ fn main() {
 }
 ```
 
+To run:
+
+```bash
+cd target/debug
+mpirun -np 2 ./<project_name>
+```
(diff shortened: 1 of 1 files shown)


bsteinb:
I have updated the example in b4b649417d14c8e4ccdb759b503fc0fd0b14b6d5 and it should now run on an arbitrary number of processes, so a simple `cargo run` should suffice. Thanks for the nudge.

thyrgle:
Cool. Thanks!

Issue #6, opened 2016-02-04
Bindgen Cargo dependencies cause rsmpi build to fail


jonalmeida:
`syntex_syntax` is the dependency that fails to build and is relied on by `bindgen`. Below, is the output of a cargo build on a Linux x64 system using the MPICH implementation:

```
➜  mpi-test git:(master) ✗ cargo run
   Compiling custom_derive v0.1.4
   Compiling rustc-serialize v0.3.16
   Compiling bindgen v0.15.0
   Compiling winapi-build v0.1.1
   Compiling libc v0.2.5
   Compiling unicode-xid v0.0.3
   Compiling bitflags v0.4.0
   Compiling libc v0.1.12
   Compiling winapi v0.2.5
   Compiling gcc v0.3.21
   Compiling conv v0.3.1
/Users/jalmeida/.multirust/toolchains/stable/cargo/registry/src/github.com-88ac128001ac3a9a/libc-0.1.12/rust/src/liblibc/lib.rs:81:21: 81:39 warning: lint raw_pointer_derive has been removed: using derive with raw pointers is ok
/Users/jalmeida/.multirust/toolchains/stable/cargo/registry/src/github.com-88ac128001ac3a9a/libc-0.1.12/rust/src/liblibc/lib.rs:81 #![allow(bad_style, raw_pointer_derive)]
                                                                                                                                                       ^~~~~~~~~~~~~~~~~~
   Compiling kernel32-sys v0.2.1
   Compiling log v0.3.5
   Compiling term v0.4.0
   Compiling syntex_syntax v0.7.0
<rustc_bitflags macros>:8:29: 8:43 error: failed to resolve. Maybe a missing `extern crate rustc_bitflags`? [E0433]
<rustc_bitflags macros>:8 bits : $ value } ; ) + impl $ crate:: __core:: fmt:: Debug for $ BitFlags {
                                                      ^~~~~~~~~~~~~~
<rustc_bitflags macros>:91:1: 93:63 note: in this expansion of bitflags! (defined in <rustc_bitflags macros>)
/Users/jalmeida/.multirust/toolchains/stable/cargo/registry/src/github.com-88ac128001ac3a9a/syntex_syntax-0.7.0/src/parse/parser.rs:88:1: 93:2 note: in this expansion of bitflags! (defined in <rustc_bitflags macros>)
<rustc_bitflags macros>:8:29: 8:43 help: run `rustc --explain E0433` to see a detailed explanation
[...]

PR #7, opened 2016-02-04
Use c_void in std::os::raw instead of libc

jonalmeida:
Only change required once #6 is fixed.

Tested with example code.


bsteinb:
Thanks for the PR. While this is the only change necessary, it turns out that all FFI type definitions are in `std::os::raw` and `libc` is not necessary as a direct dependency any more. See https://github.com/bsteinb/rsmpi/commit/269ac6d0f8a836a0ff1fcfa856aa00abbbd47621

PR #9, opened 2016-03-01
Fix links in README.md

fkjogu:
Fixes links to C wrapper.


bsteinb:
Thanks! I hate it when things silently rot like this.

PR #11, opened 2016-10-01
Mark use of extern static variables unsafe

raviqqe:
Use of extern static variables will be treated as unsafe code by rustc in the future.
This change makes rsmpi able to be compiled on the nightly as of Oct/1, 2016.

Edit: Sorry. This PR sucks because it makes rsmpi unable to be compiled on the stable due to unnecessary unsafes. Please just close it or keep it for the future.

### DIFF
--- a/src/collective.rs
+++ b/src/collective.rs
@@ -1293,7 +1293,7 @@ macro_rules! system_operation_constructors {
     ($($ctor:ident => $val:path),*) => (
         $(pub fn $ctor() -> SystemOperation {
             //! A built-in operation
-            SystemOperation($val)
+            SystemOperation(unsafe_extern_static!($val))
         })*
     )
 }
--- a/src/datatype.rs
+++ b/src/datatype.rs
@@ -93,7 +93,9 @@ macro_rules! equivalent_system_datatype {
     ($rstype:path, $mpitype:path) => (
         unsafe impl Equivalence for $rstype {
             type Out = SystemDatatype;
-            fn equivalent_datatype() -> Self::Out { SystemDatatype($mpitype) }
+            fn equivalent_datatype() -> Self::Out {
+                SystemDatatype(unsafe_extern_static!($mpitype))
+            }
         }
     )
 }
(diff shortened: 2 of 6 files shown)


bsteinb:
Well, I do not think it sucks. It is a step in the right direction, but it breaks on some configurations. That is exactly why I have this battery of CI tests set up. Let the machines help you keep track of things.

I have not investigated in detail, but I can think of two ways to move forward here:
- build a macro that conditionally (based on the compiler version) marks the use of an extern static as unsafe,
- make these variables into functions (`RSMPI_COMM_WORLD` becomes `rsmpi_comm_world()`) since they have always been unsafe.

Would you be willing to put more work into this or should I take a stab at it when I find the time?

raviqqe:
I could not find any feature to compile codes conditionally on `rustc` versions .
But, I found a way to prevent the errors which is adding:
```
#![cfg_attr(not(feature="nightly"), allow(unused_unsafe))]
```
in `lib.rs` while this seems not to be the correct way.

raviqqe:
The compilation on nightly is failing because of `clippy`'s warnings of [useless_transmute](https://github.com/Manishearth/rust-clippy/wiki#useless_transmute).
This cannot be fixed easily keeping code readable because the solution presented by `clippy` is also warned.

When `a: &T`,
```
let b: *const c_void = mem::transmute(a)
``` 
causes a warning of `useless_transmute` and it suggests:
```
a as *const T as *const c_void
```
And, this also causes a warning of `trivial_casts` on `a as *const T`.
A straightforward solution is:
```
let b: *const T = a;
b as *const c_void
```
Is this OK?

bsteinb:
I will see if I can do something about the conditional compilation.

bsteinb:
Yeah that is okay. The nightly builds are marked non-critical for that reason. I would rather have them taken care of in a separate PR.

bsteinb:
I have just pushed a macro that marks an expression as unsafe on 1.13.0 and onward. Could you rework this PR to use that macro instead of `unsafe`?

raviqqe:
Please wait a minute.

raviqqe:
That macro is so cool.

bsteinb:
Thanks for this.

PR #14, opened 2016-10-01
Remove extra CI build command options and features in Cargo.toml

raviqqe:
This PR makes build on nightly with `clippy` successful.

While I'm not sure, `cargo clippy --features clippy` seems to choose empty feature `clippy = []` in `Cargo.toml` and then cause `unknown_lints` warnings. I don't know why the original author needed to make the `lint` feature in `Cargo.toml`, which is not used anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ install:
 script:
   - cargo build -v && cargo test -v && sh ci/run-examples.sh
   - if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
-      cargo clippy -v --features "clippy";
+      cargo clippy -v;
     fi
 
 before_deploy:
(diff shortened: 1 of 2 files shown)


bsteinb:
The feature is not unused. It is set when the code is being inspected by clippy. It conditionally turns on additional lints provided by clippy, see: https://github.com/bsteinb/rsmpi/blob/57d351543f62477f4142331a57d30049424834cf/src/lib.rs#L16

This used to work but now might be broken. I will investigate later.

raviqqe:
[This](https://github.com/Manishearth/rust-clippy/pull/996) may be related to the problem.

bsteinb:
The problem has been reported here: https://github.com/Manishearth/rust-clippy/issues/1266

PR #18, opened 2017-03-27
Simplify and make Request safer

### LINKED ISSUE #17: *Request unsafe in presence of memory leaks
~~~rust
{
    let buffer = vec![0u8; 512];
    let request = dest.immediate_send(&buffer as &[u8]);
    std::mem::forget(request);
}
// request is still alive but buf is dead
~~~

a.k.a. the [JoinGuard problem](https://github.com/rust-lang/rust/issues/24292).

This could perhaps be fixed by using a [Scope-based mechanism](https://aturon.github.io/crossbeam-doc/crossbeam/struct.Scope.html), or by asking for an owned buffer instead.

Rufflewind:
Created two kinds of `Scope` objects:

  - `LocalScope` to track `Request`s with arbitrary lifetime `'a`
  - `StaticScope` (no bookkeeping, but only `'static` `Request`s)

The `Request` trait has been squashed into a single `Request<S>` data type, where `S` is the type of scope that this `Request` is associated with.

Fixes #17.

---

  - [x] Generalize methods over `Scope`.
  - [x] There's no way to transfer from an outer `LocalScope` to an inner `LocalScope`.
  - ~~It would be nice to generalize `Buffer` to allow things like `Rc`, which would be compatible with `StaticScope`.~~ (WIP: https://github.com/bsteinb/rsmpi/compare/Rufflewind:bindgen-hack...Rufflewind:buffer)

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -9,14 +9,15 @@
 //! **Note:** If the `Request` is dropped (as opposed to calling `wait` or `test` explicitly), the
 //! program will panic.
 //!
-//! To enforce this rule, every request object must be attached to some pre-existing
-//! [`Scope`](trait.Scope.html), which is usually created with [`scope`](fn.scope.html).  At the end
-//! of a `Scope`, all its remaining requests will be waited for until completion.
+//! To enforce this rule, every request object must be registered to some pre-existing
+//! [`Scope`](trait.Scope.html).  At the end of a `Scope`, all its remaining requests will be waited
+//! for until completion.  Scopes can be created using either [`scope`](fn.scope.html) or
+//! [`StaticScope`](struct.StaticScope.html).
 //!
 //! To handle request completion in an RAII style, a request can be wrapped in either
 //! [`WaitGuard`](struct.WaitGuard.html) or [`CancelGuard`](struct.CancelGuard.html), which will
 //! follow the respective policy for completing the operation.  When the guard is dropped, the
-//! request will be automatically detached from its `Scope`.
+//! request will be automatically unregistered from its `Scope`.
 //!
 //! # Unfinished features
 //!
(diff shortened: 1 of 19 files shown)


Rufflewind:
Some questions on design:

- What should the semantics of dropping a Request be?  Panic or wait?  In 6cbe9d5a046facad60118dd85f11ae9e5cc65e28 I actually did neither: I simply allowed it to be handled by the Scope, but this would be unsafe if owned buffers are attached a request (if we ever move toward that design).
- Similarly, what should be the semantics of forgetting a Request be?  Panic or wait?  In 6cbe9d5a046facad60118dd85f11ae9e5cc65e28 I chose to wait.  If we instead panicked, then the implementation of LocalScope could be simplified to just a reference counter instead of a HashMap.

Rufflewind:
Changes since yesterday:

- Removed `InactiveRequest`.
- Make `Request` panic on drop (restore old behavior for now).
- Refactored `request.rs` a bit (hopefully I didn't forget to `unregister` things this time?)
- Added `'a` to `Scope` trait, which allows …
- … generalization of `point_to_point.rs` and `collective.rs` methods over `Scope`.

Rufflewind:
So I started working on a [separate branch](https://github.com/Rufflewind/rsmpi/compare/bindgen-hack...Rufflewind:buffer) to generalize buffer types to support `'static` buffers like `Vec` or `Rc`.  That way, they can use `StaticScope` and `Vec<f64>` and not have to worry about `LocalScope`s. It's a WIP, and the names are not finalized yet (in particular, it's intended to replace the original `Buffer` and `BufferMut` traits).  Does this look interesting to you, or should I abandon this idea?

bsteinb:
Regarding the questions about the semantics of dropping or forgetting requests:

- I think the semantics for dropping can remain the same for now, so panic. And then provide the `WaitGuard` and `CancelGuard` to opt into different behavior.
- As for the semantics of forgotten requests that have not yet been deregistered when exiting from a scope: maybe it would make sense to mirror the choice of cleanup policy that is provided by the `WaitGuard` etc. here by parameterising `Scope` over another set of types `WaitForgotten`, `CancelForgotten` and `PanicForgotten`.

I am starting to wonder whether it would make sense to have the requests be owned by the scope. This would probably simplify all this cleaning up, but I think it would take away the ability to enforce a certain order of request completion.

bsteinb:
I must admit I am a bit confused by the refactoring in requests.rs. The former design where `WaitGuard` and `CancelGuard` were just wrappers around `Request` that implemented a different `Drop` and all of the logic about `wait()`ing and `cancel()`ing remained in `Request` seemed simpler. Is there a fundamental reason why this had to change that I am not seeing right now?

bsteinb:
Your other work definitely sounds interesting and I am grateful that you are investing time into this project. For the moment, I would like to focus on this PR first and move to the generalisation of buffers afterwards, if you have the patience. I am currently also juggling a regular day job, a move and a newborn, so time and mental space is scarce.

Rufflewind:
> As for the semantics of forgotten requests that have not yet been deregistered when exiting from a scope …

IMO I think parametrizing `LocalScope` would add a lot of extra complexity for what is supposed to be just a safety net.  It might be better to stick to panicking:

 - It's consistent with `Request::drop`.
 - It allows for a simple and efficient implementation of `Scope` using an integer counter (vs a full `HashMap` currently).
 - `LocalScope::drop` should only trigger if the user `forget`s a `Request` or accidentally builds a cycle with `Rc`, so panicking is a way to tell the user to fix their code.  This is similar to how `RefCell` panics if you `borrow_mut()` twice – it's a run-time safety check.

> Is there a fundamental reason why this had to change that I am not seeing right now?

When I tried it, it seemed that `drop` is called first on the outer wrapper, and then on the inner objects.  This means if I do `struct WaitRequest(Request)`, the `Request` destructor will be unconditionally called after `WaitRequest`'s destructor.

bsteinb:
> When I tried it, it seemed that drop is called first on the outer wrapper, and then on the inner objects. This means if I do `struct WaitRequest(Request)`, the `Request` destructor will be unconditionally called after `WaitRequest`'s destructor.

True, this is why `WaitGuard` etc. in the original design are `struct WaitGuard(Option<Request>)` and `drop()` `.take()`s the value out of the `Option` to perform a wait or cancel on it. This way, by the time the `drop()` of the `Option<Request>` is invoked, it contains `None` and is reduced to a no-op.

It may be somewhat redundant to wrap the `Request` in an `Option`, since the underlying raw type `MPI_Request` has a defined `null` value, but I would say this is an optimisation.

Rufflewind:
Are there any changes that need to be made to complete this PR?

bsteinb:
Not necessarily. I would still like to see the `wait()` and `cancel()` logic moved out of the `Guard`s and back into `Request` as I described in my previous comments, but I would be willing to just merge this and then work on that myself, if it is alright with you. I was just reluctant to just merge your work and then go and rewrite it.

Rufflewind:
Okay I’ve moved the logic into Request.  I originally [tried putting it in RawRequest](https://github.com/bsteinb/rsmpi/pull/18/commits/9ed7ae479db233f568d4411b5f9044c163357c04) to avoid the Option but that just made other parts more complicated.  You can squash/rebase them if you want.

Having `drop(&mut self)` [instead of `drop(self)`](https://github.com/rust-lang/rust/issues/4330) made this whole thing really awkward.

bsteinb:
Wow, again, thank you so much. Looking again at my original implementation, I notice that I misremembered and my `Guard`s did not actually use the `wait()` and `test()` methods of the `Request` trait, so thanks for going the extra mile there.

I also noticed just now, that my implementation of `cancel()` was completely unsafe even without `forget()`ting requests. Once more, good catch.

I have one more question which I will leave as an in-line comment immediately.

> Having `drop(&mut self)` instead of `drop(self)` (rust-lang/rust#4330) made this whole thing really awkward.

Yes, I remember having the exact same reaction. It reminds me of the C++ move semantics. "Yep, this value is yours. If you want to trash it, go ahead. What? No, you don't actually get to have the value. Here, have this reference instead."

bsteinb:
Ugh, apparently I had not figured out how these review things work. I left the in-line comment yesterday, but only noticed today that I needed to submit the whole review. Sorry about that.

src/request.rs
```diff
@@ -28,16 +38,94 @@ use ffi::{MPI_Request, MPI_Status};
...
+        let scope = mem::replace(&mut self.scope, mem::uninitialized());
+        let _ = mem::replace(&mut self.phantom, mem::uninitialized());
+        mem::forget(self);
+        scope.unregister();
+        (request, scope)
+    }
+
+    /// Wait for the request to finish and unregister the request object from its scope.
+    ///
+    /// This is unsafe because afterward the request is left in an unspecified but droppable state.
```
Rufflewind: Yeah this was from back when it took `&mut self` instead of `self`.

Rufflewind:
Should be fixed now!

bsteinb:
Great, let's get this merged!

PR #23, opened 2017-05-24
Make abort return !

Rufflewind:
The `!` type informs Rust that any code that follows is unreachable, which is useful because the user is no longer obligated to return anything afterward.  To do this, `process::abort()` was added to `MPI_Abort`.  It also makes sure the program dies even if `MPI_Abort` doesn’t kill the program (which shouldn't ever happen).


bsteinb:
Looks good.

PR #25, opened 2017-05-25
Add support for user-defined operations

Rufflewind:
It’s not pretty …  Because of how MPI_User_function is defined, it doesn’t give us a clean way attach/access arbitrary user data.  As a result, we have to pull in libffi to hack our way through.

- Add `BorrowedDatatype`, a thin wrapper over `MPI_Datatype` (feel free to bikeshed on the name)
- Redefine `SystemDatatype = BorrowedDatatype<'static>`
- Add `DynBuffer[Mut]`
- Add `is_commutative`
- Add `UserOperation`

### DIFF
--- a/src/collective.rs
+++ b/src/collective.rs
@@ -1515,14 +1515,19 @@ impl<'a> UserOperation<'a> {
 }
 
 unsafe fn wrapper<F>(function: &F,
-                     invec: *mut c_void,
-                     inoutvec: *mut c_void,
+                     mut invec: *mut c_void,
+                     mut inoutvec: *mut c_void,
                      len: *mut c_int,
                      datatype: *mut ffi::MPI_Datatype)
     where F: Fn(DynBuffer, DynBufferMut)
 {
     let len = *len;
     let datatype = DatatypeRef::from_raw(*datatype);
+    if len == 0 {
+        // precautionary measure: ensure pointers are not null
+        invec = [].as_mut_ptr();
+        inoutvec = [].as_mut_ptr();
+    }
     function(DynBuffer::from_raw(invec, len, datatype),
              DynBufferMut::from_raw(inoutvec, len, datatype));
 }
--- a/src/collective.rs
+++ b/src/collective.rs
@@ -6,7 +6,6 @@
 //!
 //! - **5.8**: All-to-all, `MPI_Alltoallw()`
 //! - **5.9**: Global reduction operations, `MPI_Op_create()`, `MPI_Op_free()`,
-//! `MPI_Op_commutative()`
 //! - **5.10**: Reduce-scatter, `MPI_Reduce_scatter()`
 //! - **5.12**: Nonblocking collective operations,
 //! `MPI_Ialltoallw()`, `MPI_Ireduce_scatter()`
@@ -1321,7 +1320,20 @@ impl<'a, C: 'a + Communicator> Root for Process<'a, C> {
 }
 
 /// An operation to be used in a reduction or scan type operation, e.g. `MPI_SUM`
-pub trait Operation: AsRaw<Raw = MPI_Op> { }
+pub trait Operation: AsRaw<Raw = MPI_Op> {
+    /// Returns whether the operation is commutative.
+    ///
+    /// # Standard section(s)
+    ///
+    /// 5.9.7
+    fn is_commutative(&self) -> bool {
+        unsafe {
+            let mut commute = 0;
+            ffi::MPI_Op_commutative(self.as_raw(), &mut commute);
+            commute != 0
+        }
+    }
+}
 impl<'a, T: 'a + Operation> Operation for &'a T {}
 
 /// A built-in operation like `MPI_SUM`
--- a/src/collective.rs
+++ b/src/collective.rs
@@ -5,16 +5,19 @@
 //! # Unfinished features
 //!
 //! - **5.8**: All-to-all, `MPI_Alltoallw()`
-//! - **5.9**: Global reduction operations, `MPI_Op_create()`, `MPI_Op_free()`,
 //! - **5.10**: Reduce-scatter, `MPI_Reduce_scatter()`
 //! - **5.12**: Nonblocking collective operations,
 //! `MPI_Ialltoallw()`, `MPI_Ireduce_scatter()`
 
-use std::{mem, ptr};
+use std::{fmt, mem, ptr};
+use std::os::raw::{c_int, c_void};
+
+use libffi::high::Closure4;
 
 use ffi;
 use ffi::{MPI_Request, MPI_Op};
 
+use datatype::{DatatypeRef, DynBuffer, DynBufferMut};
 use datatype::traits::*;
 use raw::traits::*;
 use request::{Request, Scope, StaticScope};
@@ -170,7 +173,7 @@ pub trait CommunicatorCollectives: Communicator {
     /// # Standard section(s)
     ///
     /// 5.9.6
-    fn all_reduce_into<S: ?Sized, R: ?Sized, O>(&self, sendbuf: &S, recvbuf: &mut R, op: &O)
+    fn all_reduce_into<S: ?Sized, R: ?Sized, O>(&self, sendbuf: &S, recvbuf: &mut R, op: O)
         where S: Buffer,
               R: BufferMut,
               O: Operation
@@ -199,7 +202,7 @@ pub trait CommunicatorCollectives: Communicator {
     fn reduce_scatter_block_into<S: ?Sized, R: ?Sized, O>(&self,
                                                           sendbuf: &S,
                                                           recvbuf: &mut R,
-                                                          op: &O)
+                                                          op: O)
         where S: Buffer,
               R: BufferMut,
               O: Operation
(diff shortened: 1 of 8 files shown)


bsteinb:
Thanks. I will try to find some time to look at this tonight.

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+        self.count().value_as().expect("Length of DynBuffer cannot be expressed as a usize")
+    }
+
+    /// Returns the underlying raw pointer.
+    pub fn as_ptr(&self) -> *const c_void {
+        self.ptr
+    }
+
+    /// Reborrow the buffer with a shorter lifetime.
+    pub fn as_ref<'b>(self) -> DynBuffer<'b> where 'a: 'b {
```
bsteinb: Better not step on `AsRef`'s toes. Why not call this `reborrow`?

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+        unsafe { DynBuffer::from_raw(self.as_ptr(),
+                                     self.count(),
+                                     self.as_datatype()) }
+    }
+}
+
+/// An mutable dynamically-typed buffer.
```
bsteinb: **A** mutable...

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
 unsafe impl<T> BufferMut for [T] where T: Equivalence
 {}
 
+/// An immutable dynamically-typed buffer.
+///
+/// The buffer has a definite length and MPI datatype, but it is not yet known which Rust type it
+/// corresponds to.  This is like the MPI analogue of `&Any`.
```
bsteinb: "like" and "analogue" in the same sentence feels like too much.

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+        unsafe { DynBuffer::from_raw(self.as_ptr(),
+                                     self.count(),
+                                     self.as_datatype()) }
+    }
+}
+
+/// An mutable dynamically-typed buffer.
+///
+/// The buffer has a definite length and MPI datatype, but it is not yet known which Rust type it
+/// corresponds to.  This is like the MPI analogue of `&mut Any`.
```
bsteinb: See above.

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+unsafe impl<'a> AsDatatype for DynBufferMut<'a> {
+    type Out = BorrowedDatatype<'a>;
+    fn as_datatype(&self) -> Self::Out {
+        self.datatype
+    }
+}
+
+impl<'a> DynBufferMut<'a> {
+    /// Creates a mutable buffer from a mutable slice with whose type has an MPI equivalent.
+    pub fn new_mut<T: Equivalence>(buf: &'a mut [T]) -> Self {
```
bsteinb: Why not just `new`?

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+    pub fn downcast_mut<T: Equivalence>(&mut self) -> Option<&mut [T]> {
+        if self.is::<T>() {
+            unsafe { Some(slice::from_raw_parts_mut(self.as_mut_ptr() as _, self.len())) }
+        } else {
+            None
+        }
+    }
+
+    /// Creates a buffer from its raw components.  The buffer must remain valid for `'a`.
+    pub unsafe fn from_raw_mut<T>(ptr: *mut T,
```
bsteinb: Why not just `from_raw`?

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+        self.ptr
+    }
+
+    /// Returns the underlying raw pointer.
+    pub fn as_mut_ptr(&mut self) -> *mut c_void {
+        self.ptr
+    }
+
+    /// Reborrows the buffer with a shorter lifetime.
+    pub fn as_ref<'b>(&'a self) -> DynBuffer<'b> where 'a: 'b {
```
bsteinb: See above.

src/datatype.rs
```diff
@@ -433,6 +450,201 @@ unsafe impl<T> BufferMut for T where T: Equivalence
...
+        unsafe { DynBuffer::from_raw(self.as_ptr(),
+                                     self.count(),
+                                     self.as_datatype()) }
+    }
+
+    /// Reborrows the buffer mutably with a shorter lifetime.
+    pub fn as_mut<'b>(&'a mut self) -> DynBufferMut<'b> where 'a: 'b {
```
bsteinb: See above.

src/collective.rs
```diff
@@ -1369,6 +1384,111 @@ unsafe impl AsRaw for SystemOperation {
...
+            function: move |invec, inoutvec, len, datatype| {
+                unsafe {
+                    wrapper(&mut function, invec, inoutvec, len, datatype)
+                }
+            },
+            wrapper: None,
+        });
+        let mut op;
+        anchor.wrapper = Some(unsafe {
+            let wrapper = ClosureMut4::new(&mut anchor.function);
```
bsteinb: There is something at the back of my mind about not `panic!`ing through the FFI. Does `libffi` take care of that or should this be handled here?

src/collective.rs
```diff
@@ -1369,6 +1384,111 @@ unsafe impl AsRaw for SystemOperation {
...
+            op = mem::uninitialized();
+            ffi::MPI_Op_create(Some(*wrapper.code_ptr()),
+                               commute as _,
```
bsteinb: Wow, I did not know this was possible. Not sure I like it, weird mix of explicit cast and inferred type. No need to change this though.

src/collective.rs
```diff
@@ -1369,6 +1384,111 @@ unsafe impl AsRaw for SystemOperation {
...
+/// The lifetime `'a` of the operation is limited by the lifetime of the underlying closure.
+///
+/// **Note:** When a `UserOperation` is passed to a non-blocking API call, it must outlive the
+/// completion of the request.  This is normally enforced by the safe API, so this is only a concern
+/// if you use the unsafe API.  Do not rely on MPI's internal reference-counting here, because once
+/// `UserOperation` is destroyed, the closure object will be deallocated even if the `MPI_Op` handle
+/// is still alive due to outstanding references.
+pub struct UserOperation<'a> {
+    op: MPI_Op,
+    anchor: Box<Erased + 'a>,           // keeps the internal data alive
```
Rufflewind: Well, the closure must be boxed (it has to be immovable so that references to it remain valid). Therefore I figured I may as well simplify the type of UserOperation by erasing F, which would've otherwise been​ an unnameable closure type.

src/collective.rs
Rufflewind: Since this is the "safe" interface of libffi, it will abort the program if the callback panics.

src/datatype.rs
```diff
@@ -60,22 +61,38 @@ pub mod traits {
...
-///
-/// # Standard section(s)
+/// A borrowed datatype handle.
 ///
-/// 3.2.2
-#[derive(Copy, Clone)]
-pub struct SystemDatatype(MPI_Datatype);
+/// This is a almost equivalent to a raw datatype handle, but unlike a raw handle, it is guaranteed
+/// to be a valid handle (subject to the lifetime `'a`).
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
```
Rufflewind: Ok, `Eq` and `PartialEq` have been removed.  (`Hash` was also removed because `Hash` by itself isn't very useful.)

src/datatype.rs
```diff
@@ -60,22 +61,38 @@ pub mod traits {
...
+pub struct BorrowedDatatype<'a> {
```
Rufflewind: I originally picked `DatatypeRef`, then swapped to `BorrowedDatatype` just before submitting the PR, but now I went back to `DatatypeRef` because `BorrowedDatatype` is too long to type :p  I would say `DatatypeRef` is a bit more consistent with e.g. [`std::cell::Ref`](https://doc.rust-lang.org/std/cell/struct.Ref.html).

Let me know if you want to go back to `BorrowedDatatype`.

src/datatype.rs
Rufflewind: Dunno, I was debating this myself.  If this were rewritten using custom dynamically sized types (https://github.com/rust-lang/rfcs/issues/813) in 2030, it would be like `&BufferObj` and `&mut BufferObj`, in which case the `_mut` suffix would be necessary to distinguish the different variants.  But I guess in the meantime writing `DynBufferMut::from_mut` feels a bit redundant.

Rufflewind:
- Added examples for `UserOperation`.
- Added `UserOperation::associative` and `UserOperation::commutative` as more readable aliases for `UserOperation::new(false/true, …)`
- The order of arguments for `UserOperation::new` has been changed to put the closure last.
- `downcast_mut` was renamed `downcast` and the old `downcast` has been removed (it's equivalent to `.downgrade().downcast()`)
- `downcast` takes `self` by value now and returns a slice with the same lifetime rather than a shorter lifetime

bsteinb:
Okay, here we go.

Rufflewind:
Cool, thanks!  Would you mind updating gh-pages?

bsteinb:
Heh, I don't even remember how to do it manually. I have Travis set-up to do it for me every time I tag a release. I guess it is time for a new release and I could even start pushing to crates.io again, now that there are no longer any dependencies on raw git repositories. I will do it by the end of the weekend.

bsteinb:
There we go. That was more of a struggle than I expected. Since today is a public holiday where I live, I will still consider this "by the end of the weekend".

Rufflewind:
Thanks! :P

PR #26, opened 2017-10-16
Be explicit that these are the sections of the MPI standard

marmistrz:
This was not clear to me, while reading the docs, TBH.
I'd ideally keep it as:
```
    /// # MPI standard reference
    /// Section 6.8, see the `MPI_Comm_set_name` function
```
but it's not trivial to automate.


bsteinb:
No, I think these are clear enough the way they are.

PR #27, opened 2017-10-18
[WIP] Implement MPI_Comm_{get,set}_name

marmistrz:
* [x] implement functions
* [x] create tests

### DIFF
--- a/examples/comm_name.rs
+++ b/examples/comm_name.rs
@@ -1,0 +1,14 @@
+#![deny(warnings)]
+extern crate mpi;
+
+use mpi::traits::*;
+
+const CNAME: &'static str = "__rsmpi__test";
+
+fn main() {
+    let universe = mpi::initialize().unwrap();
+    let world = universe.world();
+    assert_eq!("MPI_COMM_WORLD", world.get_name());
+    world.set_name(CNAME);
+    assert_eq!(CNAME, world.get_name());
+}
--- a/src/topology.rs
+++ b/src/topology.rs
@@ -20,7 +20,8 @@
 //! - **7**: Process topologies
 //! - **Parts of sections**: 8, 10, 12
 use std::{mem, process};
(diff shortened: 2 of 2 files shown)


bsteinb:
Well, have fun with your t-shirt, I guess.

marmistrz:
Hi, I actually want to get this finished but I simply didn't have the time. A pretty busy time at the university this semester. I'll comment here when I have the tests implemented.

bsteinb:
All right. It has been a while since I was a student at university and I guess I have already forgotten how hectic it gets. Sorry about that.

marmistrz:
This should be ready. I'm not doing any extra error handling, since an MPI program will crash anyway when an error is encountered. 

Let me know if this should be changed.

bsteinb:
The error handling is fine for now. I will probably want to change it once things start moving in the direction of #32, but that could take a while.

Could I ask you to change the headings for the MPI standard reference in the documentation comments to be adapted to the way they are done in the rest of the project (`# Standard section(s)` instead of `# MPI standard reference`)?

marmistrz:
The issue with `Standard section(s)` is that `standard` in this context looks like an adjective (a section, which is standard) than a noun (a section from the standard). 

`MPI standard reference` or `MPI standard section` would be much more clear to people who don't have thorough experience with MPI (most people I know who use MPI have never touched the standard and have learned from various tutorials)

bsteinb:
All right, we have discussed this issue before. My point of view has not changed and it seems neither has yours. I do not think though that either side gains anything by formatting these references in a different way on two methods out of the whole API.

bsteinb:
Thanks for the contribution.

marmistrz:
A middle ground could be to explain this in the README. What do you think?

bsteinb:
Sure, I can get behind that. Then I would also add it to the crate level documentation though.

Issue #29, opened 2018-01-30
Tracking Issue: Multiple Completion of Requests


AndrewGaspar:
Just putting some notes here on some observations from using this library -

It's kind of a pain point right now that there's no MPI_Waitall equivalent in this library. It's not too bad since you can always use mpi::ffi, but the interop experience with the existing mpi::request::Request object is poor. MPI_Waitall expects a contiguous array of MPI_Request objects, but the definition of Request makes it impossible to satisfy this requirement without allocating a separate slice of MPI_Request objects, since each MPI_Request is separated by a Scope object.

The easiest and most naive option would be to just add a wait_all operation that takes a slice (or perhaps an Iterator) of mpi::request::Request objects. It would have to internally move the MPI_Request handles into a contiguous array. APIs like MPI_Waitsome or MPI_Waitany would have some additional indirection, since they must invalidate _some_ of the MPI_Request objects that are passed in, but not all of them. You would need to pass in a `&mut [Option<mpi::request::Request>]`, and then wait_some or wait_any would need to update the slice to remove any request objects that have been completed.

I don't love this option since it imposes overhead in the name of keeping the API simple, especially for any calls that are likely to be called many times on the same set of requests such as MPI_Waitsome, MPI_Waitany, MPI_Testsome, and MPI_Testany.

The second option is to add some sort of RequestCollection object. For example:

```rust
pub struct RequestCollection<'a, S: Scope<'a> = StaticScope> {
    requests: Vec<MPI_Request>,
    scope: S,
    phantom: PhantomData<Cell<&'a ()>>,
}
```

This would allow the layout of the request objects to be directly composable with existing MPI_ multiple completion operations, while still tracking the request objects via the Scope.

This type would ideally have these features:
* It can be constructed from an iterator of mpi::request::Request objects, or via a `.push(request)` operation. This allows all existing APIs to be used to construct mpi::request::Request objects that will then be added to the collection.
* It would have member routines such as `wait_all`, `wait_some`, `test_some`, etc.
* It would support stable indexing.
* It likely would have to return `Option<Request>` rather than `Request` when indexing since wait_* may invalidate request objects.

bsteinb:
Hey there! Thanks for your input. I agree that the plural versions of the request completion routines are a very important part of the non-blocking machinery. I have not gotten around to implementing them yet, among other things because I was also still pondering possible ways of making them more idiomatic in Rust.

[...]

PR #36, opened 2018-02-11
WIP: Add support for multiple request completion

AndrewGaspar:
This pull request adds the following features to address issue #29:
* Low-level, safe multiple completion routines to the `raw` module, including support for
  * `MPI_Test`
  * `MPI_Wait`
  * `MPI_Waitany`
  * `MPI_Testany`
  * `MPI_Waitall`
  * `MPI_Testall`
  * `MPI_Waitsome`
  * `MPI_Waitany`
* A new `mpi::request::traits` module
* A new `RequestCollection` type for easily aggregating `Request` objects, with direct support for all of the above routines in multiple flavors, e.g. `wait_all_into`, `wait_all`, and `wait_all_without_status`.
  * `from_raw`, `new`, and `with_capacity` constructors
  * `push` method for adding new elements
  * `shrink` method for removing deallocated requests
  * `into_raw` method returns a Vec<MPI_Request>
  * can be constructed from an `Iterator<Item=Request>` via the `collect_requests` method on the new `CollectRequests` trait
  * Panics on drop if there are active requests in the collection
  * Does not support persistent requests
* A new `AsyncRequest` trait
  * Most `Request` functionality is moved here
* A new `Nullable` trait for MPI types that have NULL values, e.g. MPI_Request
* Three new enums
  * `mpi::raw::TestAny` for `mpi::raw::test_any`
  * `mpi::request::TestAny` for `RequestCollection::test_any` and `RequestCollection::test_any_without_status`
  * `mpi::request::TestAnyWithoutStatus` for `RequestCollection:test_any_without_status`

## TODO
- [ ] `RequestMut` struct for getting references to requests in a `RequestCollection`
- [ ] Tests for `RequestCollection::test_*` routines
- [ ] Test for `RequestCollection::test_any`
- [ ] Tests for `RequestMut` objects
- [ ] Test that `RequestCollection` panics on drop when leaking a request
- [ ] Increment minor version of rsmpi since the move from `Request` to `AsyncRequest` breaks backwards compatibility if users did not `use mpi::traits::*`


src/raw.rs
```diff
@@ -28,3 +34,324 @@ pub unsafe trait AsRawMut: AsRaw {
...
+    /// Returns the null() value of this type.
+    fn null() -> Self;
+
+    /// Returns true if this value is null. Returns false if it is not null.
+    fn is_handle_null(self) -> bool {
+        self == Self::null()
+    }
+}
+
+impl Nullable for MPI_Request {
```
bsteinb: This is fine, as long as `Nullable` is only ever implemented for `MPI_Request`, because in many MPI implementations (probably everything that is based on MPICH), all outward facing types in the MPI API (`MPI_Request`, `MPI_Comm`, ...) are aliases for `int` and thus there would be two different implementations of `Nullable` for `c_int`.

Here, we can either go with a thin newtype wrapper around `MPI_Request` or make this a simple function.

bsteinb:
Wow, thanks a lot. I already left a short comment about implementing traits for raw C API types. Should I go ahead and give this an in-depth look or do you want to complete all of the TODO items first?

src/raw.rs
AndrewGaspar: Ah, I see. I wonder if [opaque_type](https://docs.rs/bindgen/0.23.1/bindgen/struct.Builder.html#method.opaque_type) works with typedef (i.e. it will disguise the fact that it's just an int)? If so, we could use that to force unique types.

AndrewGaspar:
If you don't mind, it would be nice if you could give it a somewhat in-depth look before I proceed, just to make sure it's the right track.

src/request.rs
```diff
@@ -92,129 +201,725 @@ impl<'a, S: Scope<'a>> Request<'a, S> {
...
+    /// Converts an iterator of request objects to a `RequestCollection`. The scope of each request
+    /// must be larger than or equal of the new `RequestCollection`.
+    fn from_request_iter<'b: 'a, T, S2: Scope<'b>>(iter: T, scope: S) -> Self
+    where
+        T: IntoIterator<Item = Request<'b, S2>>,
+    {
+        let iter = iter.into_iter();
+
+        let (lbound, ubound) = iter.size_hint();
+        let capacity = if let Some(ubound) = ubound {
```
AndrewGaspar: Now that I think about this, I'm not sure this is quite right. Looking at the implementation of from_iter for Vec<_>, it seems like the correct behavior is to just use lbound, since ubound could potentially be much larger than the actual number of elements in the iterator. lbound at worst underestimates, but in what I imagine is the common case is exactly correct.

bsteinb:
All right, I will try to find time to have a look at it before the weekend.

bsteinb:
Sorry, I seem to have lost track of things. Did you want me to have a look at this or are you still working out the issues that we last discussed in #29?

AndrewGaspar:
Yeah, apologies, I've been busy with work and some personal stuff. I intend to address the buffer ownership issues before we continue reviewing.

bsteinb:
No worries. Just wanted to check if I had gotten things mixed up and you were waiting for my input.

jedbrown:
Closing in favor of #122. Thanks for laying this groundwork.

PR #46, opened 2018-09-26
Support for MS-MPI and Windows

AndrewGaspar:
This commit adds support for the MS-MPI implementation, primarily targeting Windows. I add some code to discover for MS-MPI that only run on Windows, and moved the other MPI probing to a separate probe function for Unix.

I inject an msmpi cfg into both mpi-sys and rsmpi. This may be useful to either surface msmpi specifically functionality or, like I do in this commit, to disable certain features that are missing from MS-MPI.

I noticed two issues when building with the MS-MPI headers:
1. MPI_Comm_create_group was missing.
2. Bindgen does not seem to pick up the MPI_User_function definition

For both of these issues, I just removed the APIs that called these functions when using MS-MPI. This may not be the best method.

I also disabled portions of any tests that rely on these functionalities.

I recommend building without default features on Windows since libffi is not available by default.

I also swapped the gcc crate with cc, as that seems to be the stable version of gcc.

### DIFF
--- a/mpi-sys/build.rs
+++ b/mpi-sys/build.rs
@@ -1,5 +1,5 @@
 // Compiles the `rsmpi` C shim library.
-extern crate gcc;
+extern crate cc;
 // Generates the Rust header for the C API.
 extern crate bindgen;
 // Finds out information about the MPI library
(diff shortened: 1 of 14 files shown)


mpi-sys/build.rs
```diff
@@ -26,6 +21,26 @@ fn main() {
...
+        for inc in &lib.include_paths {
+            builder.include(inc);
+        }
+
+        builder.compile("librsmpi.a");
+
+        // Adds a cfg to identify MS-MPI
+        println!("cargo:rustc-cfg=msmpi");
+    } else {
+        env::set_var("CC", "mpicc");
```
AndrewGaspar: I'm wondering if it would make sense to use the Windows methodology here for all platforms. I believe this is what, for example, FindMPI in CMake does - it doesn't actually compile your program with mpicc, It just uses the compile line from mpicc -show. This has the benefit of being able to use your preferred compiler without having to install an MPI on top of each compiler you'd like to use.

I don't know if the Windows version of this is quite robust enough, though. We'd probably need defines, too.

PR #51, opened 2018-10-10
Support packing and unpack data to byte buffers

AndrewGaspar:
Note: I'm unsure if `unpack()` needs to be unsafe. It's unclear to me that, if the typemap of the datatype is incompatible with the buffer being provided to `MPI_Unpack`, if that results in memory unsafety. I don't *think* it does, but may need another eye on this.

### DIFF
--- a/src/topology.rs
+++ b/src/topology.rs
@@ -436,7 +436,7 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
         }
     }
 
-    /// Gets the implementation-defined buffer size required to pack 'incout' elements of type
+    /// Gets the implementation-defined buffer size required to pack 'incount' elements of type
     /// 'datatype'.
     ///
     /// # Standard section(s)
--- a/src/topology.rs
+++ b/src/topology.rs
@@ -23,14 +23,16 @@ use std::ffi::{CStr, CString};
 use std::os::raw::{c_char, c_int};
 use std::{mem, process};
 
+use conv::ConvUtil;
+
+use super::Count;
 #[cfg(not(msmpi))]
 use super::Tag;
-use ffi;
-use ffi::{MPI_Comm, MPI_Group};
-
-use raw::traits::*;
 
 use datatype::traits::*;
+use ffi;
+use ffi::{MPI_Comm, MPI_Group};
+use raw::traits::*;
 
 /// Topology traits
 pub mod traits {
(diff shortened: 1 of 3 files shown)


src/topology.rs
```diff
@@ -433,6 +435,108 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
             buf_cstr.to_string_lossy().into_owned()
         }
     }
+
+    /// Gets the implementation-defined buffer size required to pack 'incout' elements of type
```
bsteinb: Should be `incount`

bsteinb:
I think `unpack` and `unpack_into` will have to be unsafe. From the description in the standard and a short experiment it does not seem as if `pack` stores any type information in the packed buffer, so `unpack` cannot do any type matching. It seems like a pair of `pack` and `unpack` can be used to the same effect as `std::mem::transmute` and so `unpack` at least should be unsafe.

AndrewGaspar:
Hm... I think you're right, but this discovery has an unfortunate side effect - I think this practically makes all communication APIs unsafe. Consider this example:

```rust
extern crate mpi;

use mpi::traits::*;

fn main() {
    let universe = mpi::initialize().unwrap();

    let comm = universe.world();

    if comm.rank() == 0 {
        comm.process_at_rank(1).send(&689i32);
    } else if comm.rank() == 1 {
        let mut float: f32 = 0.;
        comm.process_at_rank(0).receive_into(&mut float);

        println!("1: float={}", float);
    }
}
```

The output for this, at least on MS-MPI and OpenMPI 1.10.2 (what I had handy) is this:
```
GT C:\Users\andre\Projects\rsmpi [rsmpi-datatype-builder ≡ +1 ~0 -0 !]> cargo mpirun --no-default-features --example mismatched_send_recv
   Compiling mpi v0.5.4 (file:///C:/Users/andre/Projects/rsmpi)
    Finished dev [unoptimized + debuginfo] target(s) in 3.25s
1: float=0.000000000000000000000000000000000000000000965
```

I don't believe MPI actually guarantees compatible type-maps at all. We'll have to get creative to enforce true type safety.

So, am I crazy, or do we need to go through and mark any function that receives data from MPI as unsafe?

AndrewGaspar:
I'm actually not sure that specific example is "unsafe", since you can't actually produce an invalid f32 value - the type safety is preserved, and there is no memory unsafety (I believe the same is true with pack + unpack). However, you can definitely violate the type safety of `bool`:

```rust
extern crate mpi;

use mpi::traits::*;

fn main() {
    let universe = mpi::initialize().unwrap();
    let comm = universe.world();

    union BoolByte {
        b: bool,
        byte: u8,
    }

    let borb = BoolByte { byte: 2 };
    assert_ne!(false, unsafe { borb.b });
    assert_ne!(true, unsafe { borb.b });

    let mut b: bool = false;
    comm.unpack_into(&comm.pack(&2u8)[..], &mut b, 0);
    assert_ne!(false, b);
    assert_ne!(true, b);

    mpi::request::scope(|scope| {
        let process = comm.this_process();
        let request = process.immediate_send(scope, &2u8);

        let mut b: bool = false;
        process.receive_into(&mut b);
        request.wait();

        assert_ne!(false, b);
        assert_ne!(true, b);
    });
}
```

This code does not panic without a single `unsafe` block (except for the `union` part to make the point), yet it violates the type safety of `bool`. I think we have no choice to treat all receives as `unsafe`, no?

AndrewGaspar:
The existence of a APIs like [f32::from_bits](https://doc.rust-lang.org/std/primitive.f32.html#method.from_bits) and [f32::to_bits](https://doc.rust-lang.org/std/primitive.f32.html#method.to_bits) make me think that as long as every possible bit pattern in the binary representation is a valid value for the type, we're not running into any unsafety issue. I think we're only running into this issue with types with restricted values (ala `bool`). Perhaps we just want to remove the `Equivalence` impl I added for `bool`?

AndrewGaspar:
We could potentially add a `struct MpiBool(u8)` type that uses `MPI_C_BOOL` but requires explicit unwrapping to make sure the value is valid?

bsteinb:
Ugh, you are right. It appears I misread or misinterpreted the standard, since I was convinced that dynamic type checking would be performed by MPI between sending and receiving side. The information is available, but it seems it is not transmitted and nowhere in the standard does it say that that is required.

Regarding what is unsafe or not, the documentation for `transmute` cautions that it can lead to undefined behaviour. Of particular interest in our case seems to be the possibility of assigning invalid values to an object of primitive type, e.g. the `bool` problem that you describe. Any `enum` is also at risk, as are `char` and `str`. Integer and floating point types seem to be okay.

So it seems for some types (those which do not have any invalid bit-patterns) it would be okay (as in not unsafe or undefined) for Rust to get them mixed up between send and receive. The MPI standard on the other hand says that the type signature must match for typed communications (anything involving types other than `MPI_BYTE` or `MPI_PACKED`).

What to do? Do we create a trait to collect all types with no invalid bit-patterns? Have safe and unsafe variants of receive for types that are or are not members of that trait? Allow only receiving into &mut [u8] and let the user worry about serializing/deserializing?

AndrewGaspar:
After reading section 3.3 "Data Type Matching and Data Conversion", I think you're right - the MPI standard requires the program to have matching types. The exact language is:

"The type of each variable in the sender buffer has to match the type specified for that entry by the send operation; the type specified by the send operation has to match the type specified by the receive operation; and the type of each variable in the receive buffer has to match the type specified for that entry by the receive operation. A program that fails to observe these three rules is erroneous."

I think this language implies that failing to observe the type matching requirement will result in undefined behavior, which means we can't rely on the receive buffer just receiving the data bitwise. I think this requires us to mark all receive operations as `unsafe`. 😕

I'm imagining a future "safe" interface that does some sort of run-time type checking. We'd essentially have two communicator types (names are a rough draft): `UnsafeCommunicator` and `Communicator`. A couple options here:
1. Have separate unsafe communicator operations and safe communicator operations that can both operate on the same `MPI_Comm` object. You cannot pair an `unsafe` send operation with a safe receive operation, or vice-versa. To re-iterate: this requires the "unchecked" send operation to be marked as `unsafe`, as well.
2. Have two separate communicator types: one with safe send and unsafe receive operations, and one with safe send AND receive operations using runtime type checking. It is invalid to use a communicator as both an "unsafe" communicator and a safe "communicator". I.e. you must explicitly create a separate `MPI_Comm` object if you want to use the other communicator operation types. Otherwise an "unchecked" send could erroneously pair with a "checked" receive and cause undefined behavior without any `unsafe` could.

Both of these approaches should allow MPI programs with zero-`unsafe` code to be guaranteed to not exhibit MPI-introduced undefined behavior.

AndrewGaspar:
All of this to say, `unpack` should be `unsafe`, too.

Issue #12, opened 2016-10-01
Translate MPI thread support levels into Rust Send and Sync


bsteinb:
MPI defines several levels of thread support that indicate "how thread safe" the currently initialized MPI environment is. This level is only known at run-time and could be changed by re-initializing MPI (?).

In Rust, the thread safety of an API is defined by having its types implement the `Send` and `Sync` traits, which is a compile-time contract.

These traits are automatically derived by the compiler which currently leads to wrong results for `rsmpi`, e.g.:
- when building against MPICH which uses integers for its handle types, the Rust compiler always deem them safe to `Send` and `Sync`,
- when building against OpenMPI which uses pointers for its handle types, the Rust compiler will always deem them not safe to `Send` and `Sync`.

Plan: Add markers to rsmpi types so they can have variants that statically are or are not thread safe and have sources such as `world()` or `equivalent_type()` return enum values that return one of those variants depending on the current level of thread support of the environment.

Rufflewind:
Could perhaps add dummy types to represent thread-safety guarantees:

~~~rust
pub mod thread {

// !Send + !Sync
pub struct Funneled(PhantomData<*const ()>);

// Send + !Sync
pub struct Serialized(PhantomData<Cell<()>>);

// Send + Sync
pub struct Multiple(());

}

// marker trait that promises Self has correct Send/Sync semantics
pub unsafe trait Thread {
    fn raw_value() -> libc::c_int;
}

unsafe impl Thread for thread::Funneled { /* ... */ }
unsafe impl Thread for thread::Serialized { /* ... */ }
unsafe impl Thread for thread::Multiple { /* ... */ }
~~~

The rest of the API would then parametrize over these dummy types, e.g.

~~~rust
pub struct Universe<M> {
    buffer: Option<Vec<u8>>,
    thread: Phantom<M>,
}

pub enum InitializeError {
    AlreadyInitialized,
    InsufficientThreadSupport,
}

pub fn initialize<M: Thread>() -> Result<Universe<M>, InitializeError> {
    if is_initialized() {
        Err(AlreadyInitialized)
    } else {
        let mut provided: c_int = unsafe { mem::uninitialized() };
        let required = M::raw_value();
        unsafe {
            ffi::MPI_Init_thread(ptr::null_mut(),
                                 ptr::null_mut(),
                                 required,
                                 &mut provided);
        }
        if provided >= required {
            Ok(Universe { buffer: None, thread: PhantomData })
        } else {
[...]

PR #52, opened 2018-10-10
Support for #[derive(Equivalence)] to allow Rust structs to be sent directly over MPI

AndrewGaspar:
NOTE: This PR should be merged after #49 and #51.

This adds a crate that provides a custom derive for `Equivalence`, allowing a user to mark a complex struct with `#[derive(Equivalence)]` to be able to safely send that type via MPI without having to manually serialize the data.

E.g.
```rust
#[derive(Equivalence, Default)]
struct MyDataRust {
    b: bool,
    f: f64,
    i: u16,
}

let mut my_data = if world.rank() == root_rank {
    MyDataRust { b: true, f: 3.4, i: 7 }
} else {
    MyDataRust::default()
};

root_process.broadcast_into(&mut my_data);
// all ranks should have MyDataRust { b: true, f: 3.4, i: 7 }
```

This should work with any `struct` in Rust that is composed of fundamental types (or arrays of fundamental types) that implement `Equivalence`. I believe I've tested all forms of structs (with named fields, unnamed fields, and no fields).


benruijl:
This is great feature! Is there any timeline for when this can be merged?

AndrewGaspar:
Oh, I was originally hoping for a review... I should probably just update it and merge it.

fzyzcjy:
Wow this is really helpful and can reduce boilerplate! Looking forward to seeing it merged :)

AndrewGaspar:
Haha, ok, I've been convinced, let me get this done...

AndrewGaspar:
I was working on this last night - mainly the issue is that the diff was a little gnarly. However, I did run into another issue of significance - the `Equivalence` derive requires doing a bunch of "offset_of"s to get the offset of each field in the struct. The problem with this is that it requires UB to do (and was also using the deprecated `mem::uninitialized`, but switching to `MaybeUninit` doesn't fix the UB).

There's discussion of this issue in the `memoffset` crate (which we don't use, but we accomplish the task in a similar way): https://github.com/Gilnaa/memoffset/issues/24

Of note, there's an implicit promise that even though the code in `memoffset` is technically UB, the compiler team will ensure that the code continues to work as expected: https://github.com/Gilnaa/memoffset/issues/24#issuecomment-524757638

So there are two options:
- Take Ralf at his word that this pattern won't be broken by the compiler ("defined" UB)
- Just require the user to `#[derive(Default)]`, which would eliminate all UB (we can construct a default object to get the field offsets)

Any thoughts?

adamreichold:
> Any thoughts?

I would suggest deferring to the `memoffset` crate as a community-wide solution should eventually materialise there. Until it does we will have working code without a future maintenance burden. It does certainly seem worth the dependency...

src/topology.rs
```diff
@@ -433,6 +435,108 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
             buf_cstr.to_string_lossy().into_owned()
         }
     }
+
+    /// Gets the implementation-defined buffer size required to pack 'incout' elements of type
```
adamreichold, line 439: Typo `incout` instead of `incount`.

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,136 @@
+#![recursion_limit = "256"]
+
+extern crate proc_macro;
+extern crate proc_macro2;
+
+#[macro_use]
+extern crate quote;
+extern crate syn;
```
adamreichold, line 8: Have you considered using the [`synstructure`](https://crates.io/crates/synstructure) to reduce the boiler plate necessary here?

adamreichold:
I would be glad if the support MPI packing could be moved into a separate PR to make it easier to review this one.

AndrewGaspar:
> I would be glad if the support MPI packing could be moved into a separate PR to make it easier to review this one.

The diff is misleading - the packing stuff was already merged: https://github.com/rsmpi/rsmpi/pull/51

src/topology.rs
AndrewGaspar: Already fixed. :) https://github.com/rsmpi/rsmpi/pull/51/files#diff-d6bd1f3117e0d25895fecc79cf35314dR439

mpi-derive/src/lib.rs
AndrewGaspar: I can take a look. Haven't heard of it before.

AndrewGaspar:
> I would suggest deferring to the memoffset crate as a community-wide solution should eventually materialise there.

Hm... unfortunately, I can't slot in `memoffset::offset_of!` because it doesn't support tuples. I'll just switch to `std::mem::MaybeUninit` and deal with the same UB that `memoffset` has, but I'll make a note of it and reference back here.

mpi-derive/src/lib.rs
AndrewGaspar: Briefly looked into it. I'm not sure how helpful it would be. I think since the code is already written against `syn`, we should just merge it as is. You're welcome to try refactoring it if you'd like - happy to be proven wrong.

AndrewGaspar:
Tests are passing, so this can be merged as long as there's approval.

.gitignore
```diff
@@ -1,2 +1,3 @@
 Cargo.lock
 target
+.vscode/
```
adamreichold, line 3: I think editor/IDE-specific stuff like this belongs into your Git exclude file?

adamreichold:
> > I would suggest deferring to the memoffset crate as a community-wide solution should eventually materialise there.
> 
> Hm... unfortunately, I can't slot in `memoffset::offset_of!` because it doesn't support tuples. I'll just switch to `std::mem::MaybeUninit` and deal with the same UB that `memoffset` has, but I'll make a note of it and reference back here.

I understand that you just want to get going, but wouldn't adding support for tuples to `memoffset` be a better course of action? That way the whole ecosystem benefits and we do not reinvent the wheel here. (Also there could even be a reason why `memoffset` does not support them?)

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+
+            // This is very UB, but Rust core devs insist this pattern will not be broken by
+            // compiler updates. See: https://github.com/rsmpi/rsmpi/pull/52#issuecomment-697835472
+
+            let value: &#type_ident = unsafe { &*value_uninit.as_ptr() };
+
+            let value_loc = value as *const _ as usize;
+            let offset_loc = &value.#field_name as *const _ as usize;
+
+            std::mem::forget(value);
```
adamreichold, line 33: There is no need to forget `MaybeUninit<T>` as it will never call 'T'`s `Drop` impl automatically.

AndrewGaspar: Ah, yes - thought about this and forgot to remove.

.gitignore
AndrewGaspar: Hm... too bad there's not a user-level way to ignore editor files. Only seems to be on a repo-by-repo basis.

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+    let field_datatypes = fields.iter().map(equivalence_for_field);
+    let datatypes = quote! {[#(::mpi::datatype::UncommittedDatatypeRef::from(#field_datatypes)),*]};
+
+    quote! {
+        unsafe impl ::mpi::datatype::Equivalence for #ident {
+            type Out = ::mpi::datatype::DatatypeRef<'static>;
+            fn equivalent_datatype() -> Self::Out {
+                use ::mpi::raw::AsRaw;
+
+                thread_local!(static DATATYPE: ::mpi::datatype::DatatypeRef<'static> = {
```
adamreichold, line 116: Using TLS access could give deriving `Equivalence` a performance disadvantage compared to implementing it manually. If we are unable to construct the `DatatypeRef` in const context, maybe something like [`Lazy`](https://crates.io/crates/once_cell) is more appropriate even though it would still imply overhead.

Since we actually do not need exactly once semantics and the final result will be a single point AFAIU, it might be better to load that via a relaxed atomic access, compute and store the datastore the datatype (this could then happen multiple times with the same result) and then set the datatype reference. This way the common case would reduce to a single relaxed atomic load which should have no hardware overhead on e.g. x86-64.

src/lib.rs
```diff
@@ -121,6 +121,9 @@
 use std::mem::MaybeUninit;
 use std::os::raw::c_int;
 
+// Re-export derives
+pub use mpi_derive::*;
```
adamreichold, line 125: Due to the significant compile time hit currently associated with derive macros, it would be nice if this could be behind a feature flag like `derive`

AndrewGaspar:
Filed a WIP PR on memoffset: https://github.com/Gilnaa/memoffset/pull/48

src/lib.rs
AndrewGaspar: Ah yes, I also thought about that and forgot. haha

.gitignore
adamreichold: There is, just put something like
```
[core]
    excludesFile = ~/.gitignore
```
into your user-level `.gitconfig` and populate `~/.gitignore` with your tool-specific ignores.

AndrewGaspar: Neat!

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+    };
+
+    let displacements = quote! {[#(#field_displacements as ::mpi::Address),*]};
+
+    let field_datatypes = fields.iter().map(equivalence_for_field);
+    let datatypes = quote! {[#(::mpi::datatype::UncommittedDatatypeRef::from(#field_datatypes)),*]};
+
+    quote! {
+        unsafe impl ::mpi::datatype::Equivalence for #ident {
+            type Out = ::mpi::datatype::DatatypeRef<'static>;
```
adamreichold, line 112: Are there any soundness issues here if I use the derive on a struct containing lifetimes strictly smaller than `'static` yet still produce a `DatatypeRef<'static>`?

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+
+fn equivalence_for_field(field: &syn::Field) -> TokenStream2 {
+    equivalence_for_type(&field.ty)
+}
+
+fn equivalence_for_struct(ast: &syn::DeriveInput, fields: &Fields) -> TokenStream2 {
+    let ident = &ast.ident;
+
+    let field_blocklengths = fields.iter().map(|_| quote! {1 as ::mpi::Count});
+    let blocklengths = quote! {[#(#field_blocklengths),*]};
```
adamreichold, line 88: I think it would be nice if this would mirror `displacements`, e.g.
```
let field_blocklengths = fields.iter().map(|_| 1);
let blocklengths = quote! {[#(#field_blocklengths as ::mpi::Count),*]};
```
which should also reduce number of times `quote!` is called.

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+    match ty {
+        Type::Path(ref type_path) => quote!(
+                <#type_path as ::mpi::datatype::Equivalence>::equivalent_datatype()),
+        Type::Tuple(ref type_tuple) => equivalence_for_tuple_field(&type_tuple),
+        Type::Array(ref type_array) => equivalence_for_array_field(&type_array),
+        _ => panic!("Unsupported type!"),
+    }
+}
+
+fn equivalence_for_field(field: &syn::Field) -> TokenStream2 {
```
adamreichold, line 80: I think this could be a closure used directly in line 107 to reduce the call chain chasing when reading this.

mpi-derive/src/lib.rs
AndrewGaspar: Hm... I'm trying to imagine what it would even mean to apply `Equivalence` to a structure that has an associated lifetime. I can see about asserting lifetimes more. Currently `Equivalence` is not applied to any types that aren't `'static`, so you'd have to manually implement `Equivalence` to an object with a lifetime.

adamreichold: I guess for practical applications, the references must be ignored in the `Equivalence` implementations as it makes little sense putting them on the wire.

So deriving would fail as no `Equivalence` impl for any `&T` exists and here we are fine. (Except if the user adds `unsafe impl Equivalence for &'a T` in which case it is their own fault.)

adamreichold: >  I can see about asserting lifetimes more.

I don't think it is necessary to make it more complicated. My question really was a genuine question upon seeing the `'static' lifetime there.

mpi-derive/src/lib.rs
adamreichold: Thinking about this some more: Instead of a pointer to some data structure we effectively store an `MPI_DataType = i32` value which is a handle for some data structure internal to the MPI implementation, right? So we actually do want exactly once semantics to avoid creating multiple MPI-side representations of the same data structure? Also do we have a threading issue here if the MPI implementation does not support multi-threading? I guess `Lazy` would be the safest choice with the least overhead then. (And a path of being part of the standard library eventually...)

adamreichold: After thinking about this a bit more, I wonder if we can memoise at all here: If initialize MPI, use this equivalence, finalize MPI and initialize it again, and reuse this equivalence, I will end up passing a data type handling from when I first initialized the MPI stack to the second instance which might refer to a completely different data type. As it seems, we should probably always call into the MPI implementation to get the proper data type, or we should at least attach our cache of data type handles to the MPI universe? I guess not memoising but trying to cache the directly type-derived values like offset and displacement arrays would be best?

src/lib.rs
AndrewGaspar: Done.

AndrewGaspar:
Can't be merged until https://github.com/Gilnaa/memoffset/pull/48 is merged and a new release of memoffset is cut.

mpi-derive/src/lib.rs
AndrewGaspar: good catch.

mpi-derive/src/lib.rs
AndrewGaspar: >  If initialize MPI, use this equivalence, finalize MPI and initialize it again

This is illegal according to the standard.

image

We really ought to add some sort of assertion into rsmpi to ensure this since it could probably result in undefined behavior - the implementation isn't technically required to do anything that Rust considers safe.

> Also do we have a threading issue here if the MPI implementation does not support multi-threading?

I think the answer is yes - there actually needs to be some work done w.r.t. optional thread-safety in rsmpi in general. See #12 

> I guess Lazy would be the safest choice with the least overhead then.

Technically, even this has problems - if the user initializes mpi with `Threading::Single` (which is the default), then even calling into MPI from another thread, even if it's perfectly thread safe from the application's perspective, is illegal.

How about this - we use `once_cell::Lazy`, and if MPI is initialized with anything other than `Threading::Multiple`, panic if the thread that is attempting to initialize the datatype is different from the thread that called `mpi::initialize*`. That gives me an opportunity to add the check for re-initialization described before.

AndrewGaspar: Nvm, we're actually already safe from the perspective of MPI re-initialization - `mpi::initialize*` returns `None` if `mpi::is_initialized()`. So I just need to add the described check in the `Equivalence` derive.

AndrewGaspar: The latest change addresses your concerns, I believe. It correctly checks for non-`Threading::Multiple` that the thread which initializes the datatype is the one that initialized MPI. It also uses `Lazy` now rather than `thread_local`.

Cargo.toml
```diff
@@ -16,20 +16,47 @@ license = "MIT/Apache-2.0"
...
 user-operations = ["libffi"]
+derive = ["mpi-derive"]
 
 [dependencies]
 conv = "0.3"
 libffi = { version = "0.8.0", optional = true }
+mpi-derive = { path = "mpi-derive", optional = true }
 mpi-sys = { path = "mpi-sys", version = "0.2" }
+# Public dependency when "derive" feature is enabled
+once_cell = "1.4"
```
adamreichold, line 33: Can we make `once_cell` and `memoffset` optional and enable them only when `derive` is enabled?

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+                    ::mpi::datatype::internal::check_derive_equivalence_universe_state(#ident_str);
+
+                    let datatype = ::mpi::datatype::UserDatatype::structured::<
+                        ::mpi::datatype::UncommittedDatatypeRef,
+                    >(&#blocklengths, &#displacements, &#datatypes);
+
+                    let datatype_ref =
+                        unsafe { ::mpi::datatype::DatatypeRef::from_raw(datatype.as_raw()) };
+
+                    ::std::mem::forget(datatype);
```
adamreichold, line 113: Is our MSRV sufficient to use `ManuallyDrop` here?

Cargo.toml
AndrewGaspar: `memoffset` can be optional, but now `once_cell` is being used to address a race condition that would have made it technically possible to call `MPI_Init_thread` simultaneously from multiple threads.

AndrewGaspar: I might use `memoffset` to provide default `derive Equivalence` for naked tuples up to some number of elements, though.

adamreichold: I see now that `once_cell::Lazy` is always used for the `UNIVERSE_STATE`.

mpi-derive/src/lib.rs
AndrewGaspar: Yeah, the code requires `1.31` at least for edition 2018. Currently we're not testing an MSRV, so I'm actually not sure what it is.

mpi-derive/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    let result = match ast.data {
+        syn::Data::Enum(_) => panic!("#[derive(Equivalence)] is not compatible with enums"),
+        syn::Data::Union(_) => panic!("#[derive(Equivalence)] is not compatible with unions"),
+        syn::Data::Struct(ref s) => equivalence_for_struct(&ast, &s.fields),
+    };
+    result.into()
+}
+
+fn offset_of(type_ident: &dyn quote::ToTokens, field_name: &dyn quote::ToTokens) -> TokenStream2 {
+    quote!(::memoffset::offset_of!(#type_ident, #field_name))
```
adamreichold, line 21: I think this would imply that a crate using this crate has a memoffset depedency? I think we should re-export it like
```
mod internal {
pub use memoffset;
}
```
and then use our version here like
```
::mpi::internal::memoffset::*
```
?

AndrewGaspar: Oh, I haven't seen this pattern before, but makes sense.

Cargo.toml
AndrewGaspar: Yeah, and even though we (currently) don't need the state for anything but the derive feature, we still need the mutex to guard against the previously unnoticed initialization race condition.

mpi-derive/src/lib.rs
adamreichold: > The latest change addresses your concerns, I believe. It correctly checks for non-Threading::Multiple that the thread which initializes the datatype is the one that initialized MPI. It also uses Lazy now rather than thread_local.

Yes, I think so too and it is quite a tour de force! (I am especially grateful that you went out of your way to improve the ecosystem via `memoffset`. Thank you for that!)

One thing I do not yet understand is whether the `// TODO and NOTE:` still applies, at least specifically for the code in question? You do have the lock on `UNIVERSE_STATE`, doesn't this prevent races with initialisation/finalisation?

AndrewGaspar: > One thing I do not yet understand is whether the // TODO and NOTE: still applies, at least specifically for the code in question? You do have the lock on UNIVERSE_STATE, doesn't this prevent races with initialisation/finalisation?

It prevents initialization from racing with itself, and it prevents the "main thread" check from racing with initialization and finalization, but it doesn't prevent the actual implementation of `equivalent_datatype` from racing with finalization. But this issue is basically true of any API that allows you the call `MPI_` routines without a borrow on the `Universe`, which is almost the entire API. So I decided to just punt on it for now.

adamreichold: > But this issue is basically true of any API that allows you the call MPI_ routines without a borrow on the Universe, which is almost the entire API. So I decided to just punt on it for now.

Understood. Sounds reasonable. I guess we will need to add borrowing the universe to all API that creates handle-like objects eventually to fix #12, but this is certainly not the place to attack that issue.

mpi-derive/src/lib.rs
AndrewGaspar: Actually, I'll just change the static to store `UserDatatype` directly and then do the conversion to `DatatypeRef<'static>` outside the `Lazy` initialization. I don't know why I didn't do that in the first place.

AndrewGaspar:
@adamreichold Look good?

adamreichold:
> @adamreichold Look good?

I admittedly did not have time yet to look over your latest changes. Could you give me until the middle of the week? I think I will be able to schedule a more comprehensive review until then.

Cargo.toml
```diff
@@ -16,20 +16,46 @@ license = "MIT/Apache-2.0"
...
 [build-dependencies]
 rustc_version = "0.2"
+
+# The following tests depend on specific features
+
+[[example]]
+name = "struct"
+required-features = ["derive"]
+
+[[example]]
```
adamreichold, line 47: I wonder if those examples should actually be integration tests? (They might confuse rather than help new users.)

adamreichold:
> @adamreichold Look good?

@AndrewGaspar Besides the cosmetic question of whether to use examples or tests, this looks good to me now.

Cargo.toml
AndrewGaspar: Yeah, unfortunately we don't have a good setup to do this a different way right now... We don't have an easy way to run cargo tests with mpiexec. We should definitely fix this in the long term, though.

adamreichold: Setting the runner for `cfg(test)` to `mpiexec` via [Cargo's configuartion](https://doc.rust-lang.org/cargo/reference/config.html#targetcfgrunner) would not suffice due to setup/tear down issues?

PR #58, opened 2018-11-08
Adds support for Cartesian-topology communicators

AndrewGaspar:
Adds a new `CartesianCommunicator` type that is, in a way, a sub-class of `UserCommunicator`. A `CartesianCommunicator` provides an n-dimensional mapping of ranks, and can even re-order ranks to improve spatial locality.

`UserCommunicator` objects can be promoted to a more specific type representing their topology. This is currently unimplemented for graph and distributed graph communicators.

The new routines on `CartesianCommunicator` generally have both unchecked, unsafe and checked, safe variants, where the unchecked variants are labeled as such.

`topology.rs` was getting a little hairy, so I moved it to `topology/mod.rs` and added a `topology/cartesian.rs` for the Cartesian-specific code. All public names still appear under the `mpi::topology` namespace.

I also added a dependency on [smallvec](https://crates.io/crates/smallvec). This was to do conversions from the more idiomatic `&[bool]` to the arrays of `c_int` values that MPI expects for lists of boolean values without having to do a heap allocation in the vast majority of cases.


AndrewGaspar:
Rebased - CI running now.

PR #72, opened 2019-11-19
Document Spectrum MPI support

AndrewGaspar:
Tested with spectrum-mpi on Power9. One test is failing (immediate.rs) - seeing if I can understand why. But it generally works.


AndrewGaspar:
The test failure is a bug in Spectrum MPI with matched MPI probes.

bsteinb:
Wow, exciting stuff! One thing: the libraries that are currently listed here are tested continuously on Travis. Could you change the text to make clear that this is not the case (I assume) for spectrum-mpi?

AndrewGaspar:
I made it clear that it's not tested in CI.

PR #74, opened 2019-12-06
Add macro for defining scopes

AndrewGaspar:
The current mechanism for creating `LocalScope`s relies on a function and closures. This change adds a macro named `mpi::define_scope!` for defining the `LocalScope` instead, which allows the user to stay in their local function context, rather than having a separate function. This enables more rich flow control when dealing with scopes, and also unifies scope creation for when I add async support.

For example, `define_scope!` can be used like this:

```rust
// define send and receive buffers here
{
    // you may define send buffers here
    mpi::define_scope!(scope);
    // use scope like normal with receive and send buffers
}
// use receive buffers here
```

The macro works and is safe because the created `&LocalScope` shadows the stack-pinned `LocalScope`, making it impossible to forget, or even reference, the underlying `LocalScope`.

I also changed all the examples to use `mpi::define_scope!(scope);`.

There's probably a better name than `mpi::define_scope`, btw. Maybe just `mpi::scope!`? Or `mpi::request_scope!`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,37 +81,40 @@ fn main() {
     let previous_rank = if rank > 0 { rank - 1 } else { size - 1 };
 
     let msg = vec![rank, 2 * rank, 4 * rank];
-    mpi::request::scope(|scope| {
-        let _sreq = WaitGuard::from(
-            world
-                .process_at_rank(next_rank)
-                .immediate_send(scope, &msg[..]),
-        );
 
-        let (msg, status) = world.any_process().receive_vec();
+    // Defining scopes ensures that all borrowed buffers live until all requests
+    // that borrow them complete.
+    mpi::define_scope!(scope);
 
-        println!(
-            "Process {} got message {:?}.\nStatus is: {:?}",
-            rank, msg, status
-        );
-        let x = status.source_rank();
-        assert_eq!(x, previous_rank);
-        assert_eq!(vec![x, 2 * x, 4 * x], msg);
+    let _sreq = WaitGuard::from(
+        world
+            .process_at_rank(next_rank)
+            .immediate_send(scope, &msg[..]),
+    );
 
-        let root_rank = 0;
-        let root_process = world.process_at_rank(root_rank);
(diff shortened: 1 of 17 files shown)


AndrewGaspar:
Hm... defining Scope for pinned LocalScope isn't actually sound and I'm not sure why I thought it was. Marking as WIP.

I think the macro is fine, but trying to use pinning to solve the issue doesn't work.

AndrewGaspar:
Requires #84 to pass CI.

adamreichold:
To be honest, the macro-based approach is a bit too magical for my taste. Scope/closure based API are a standard idiom when lifetimes need to be controlled so this change could make `rsmpi` harder to learn. Injecting an `unsafe` block into user code could also unfavorably interact with things like `#![deny(unsafe)]` AFAIU.

AndrewGaspar:
> Injecting an unsafe block into user code could also unfavorably interact with things like #![deny(unsafe)] AFAIU.

Should be fixable with an `#[allow(unsafe)]`, right?

> To be honest, the macro-based approach is a bit too magical for my taste.

Well, I'm not super attached to this change, but the big reason I want it is for async/await (which I'd like to tackle seriously for 0.7). It would unify defining scopes for both sync code and async code. We can maybe push this off for 0.7.

adamreichold:
> Should be fixable with an #[allow(unsafe)], right?

Yes, I think so.

>  We can maybe push this off for 0.7.

I would be glad if we did as the interactions with pinning should be worked out first. Polling a write on a socket for completion basically has the same problem of borrowing a buffer and then yielding. Hence an implementation of immediate operations based on async/await might not have this problem at all.

jedbrown:
Is the main benefit here that you can have overlapping scopes? For example, in an atmospheric simulation, you can have a scope for communication with horizontal neighbors and another scope to communicate with vertical neighbors. With this macro, they can overlap, but they have to be either nested or disjoint if using a closure. I probably need to experiment.

Issue #75, opened 2019-12-06
LocalScope Drop impl is not panic safe


AndrewGaspar:
This code with no uses of `unsafe` leaks a buffer to MPI.

```rust
use mpi::traits::*;

fn main() {
    let universe = mpi::initialize().unwrap();

    let comm = universe.world();

    {
        let x = 0.0;
        std::panic::catch_unwind(|| {
            mpi::request::scope(|scope| {
                // Leaks the Request
                std::mem::forget(
                    comm.process_at_rank(comm.rank()).immediate_send(scope, &x),
                );
            });
        });
    }

    println!("Oh no, we leaked a buffer to MPI without unsafe!")
}
```

We probably need to change the `Drop` impl for `LocalScope` to a call to [`std::process::abort`](https://doc.rust-lang.org/std/process/fn.abort.html).

fzyzcjy:
Hi maybe my problem is related to this? https://github.com/rsmpi/rsmpi/issues/94 Thanks!

PR #77, opened 2019-12-08
Terminate execution when scoped requests are leaked.

AndrewGaspar:
Makes LocalScope panic safe

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -348,15 +348,21 @@ pub struct LocalScope<'a> {
     phantom: PhantomData<Cell<&'a ()>>, // Cell needed to ensure 'a is invariant
 }
 
+#[cold]
+fn abort_on_unhandled_request() {
+    let _ = std::panic::catch_unwind(|| {
+        panic!("at least one request was dropped without being completed");
+    });
+
+    // There's no way to tell MPI to release the buffers that were passed to it. Therefore
+    // we must abort execution.
+    std::process::abort();
+}
+
 impl<'a> Drop for LocalScope<'a> {
     fn drop(&mut self) {
         if self.num_requests.get() != 0 {
-            std::mem::forget(std::panic::catch_unwind(|| {
-                panic!("at least one request was dropped without being completed");
-            }));
-            // There's no way to tell MPI to release the buffers that were passed to it. Therefore
-            // we must abort execution.
-            std::process::abort();
+            abort_on_unhandled_request();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


AndrewGaspar:
Fixes #75

AndrewGaspar:
Any problems with this change? I think it's necessary for safety issue.

src/request.rs
```diff
@@ -351,7 +351,12 @@ pub struct LocalScope<'a> {
 impl<'a> Drop for LocalScope<'a> {
     fn drop(&mut self) {
         if self.num_requests.get() != 0 {
-            panic!("at least one request was dropped without being completed");
+            std::mem::forget(std::panic::catch_unwind(|| {
```
adamreichold, line 354: Why `panic!`, `catch_unwind` and `forget` instead of just `eprintln!`? Is this to produce a stack trace? And if so, why would
```
let _ = std::panic::catch_unwind(|| ...);
```
not suffice to ignore the result? Why must it be leaked?

adamreichold:
> Any problems with this change? I think it's necessary for safety issue.

Personally, I have no issue with aborting here. (I build simulation code with `panic = "abort"` in any case for the tiny bit of code density improvement it enables.)

src/request.rs
AndrewGaspar: Yeah, that's why I wanted to keep the `panic!` - to maintain context.

You're right, I probably can use `let _ = ` - I'll fix that.

adamreichold: One thing I am still considering is code density again. IIRC, some places in the standard library do plain `abort` to avoid introducing jumps into the panic machinery in hot code parts.

Considering context: `SIGABRT` should stop my debugger right at the call to `abort` and allow me to debug the issue, so personally, I could live without the panic here.

If you do keep it, it might be useful to put the how `panic`, `catch_unwind`, `abort` sequence into a separate function and mark that `#[cold]` or at least `#[inline(never)]` to avoid putting those landing pads everywhere `LocalScope`'s `Drop` implementation is inlined.

adamreichold: (Just for reference, I think I was remembering places like https://github.com/rust-lang/rust/blob/08deb863bdebfcbbb71c18acf903eca84f1df4e7/library/alloc/src/sync.rs#L1174 and https://github.com/rust-lang/rust/blob/2d6cbd21b2c5819c7fa42eb5a3713667b67e4f03/library/alloc/src/rc.rs#L2107)

adamreichold:
LGTM

PR #79, opened 2019-12-08
Require safe transmutability traits for buffers

### LINKED ISSUE #54: Communication operations, both send and receive, must be marked as unsafe
Mis-matching datatypes when performing communication results in undefined behavior.

Documentation should indicate that the `unsafe` here need not infect the whole code base. A routine that properly matches its own communications is itself safe. You only get unsafety if another routine incorrectly matches their own sends and receives. Ergo, it is _possible_ to expose a safe interface on top of MPI communications, even if the atoms are `unsafe`.

See #51 for discussion on this.

AndrewGaspar:
I added two new traits inspired by this [Pre-RFC](https://internals.rust-lang.org/t/pre-rfc-v2-safe-transmute/11431): `ToBytes` and `FromAnyBytes`. `ToBytes` can be safely applied to any type with deterministic layout that has no internal padding (and therefore has no information disclosure potential). `FromAnyBytes` can be safely applied to any type where all potential bit patterns of its components are valid instances of the type.

This change makes the default `Buffer` implementation require `ToBytes`, and the `BufferMut` implementation require `FromAnyBytes`. This fixes issues related to receiving invalid values into `bool`.

This fixes #54.


AndrewGaspar:
Related: https://internals.rust-lang.org/t/pre-rfc-v2-safe-transmute/11431

AndrewGaspar:
I hadn't considered the issue of sending structs with padding - may have to switch to require something equivalent to `ToBytes` for send buffers.

AndrewGaspar:
I'm looking to see if we can get a full "0.11" release for `safe_transmute`. https://github.com/nabijaczleweli/safe-transmute-rs/issues/62

AndrewGaspar:
I removed the public `safe-transmute` dependency (though still use it in an example) and added the desired traits myself. I updated the MR text to reflect this.

adamreichold:
> I removed the public `safe-transmute` dependency (though still use it in an example) and added the desired traits myself. I updated the MR text to reflect this.

That release seems to have happened in the mean time. I think the public dependency would actually be a better choice as interoperability w.r.t. these key traits could be automatic, e.g. other crates could implement this for their data types and we would automatically benefit. Ecosystem service effects and all that...

adamreichold:
I see two concerns here:

1. I would like to see `MaybeUninit<M>` included in the fold so that I can still receive arbitrary data types which are not safe to transmute even though I will have to make the assertion of correct communication by calling the unsafe `assume_init` in my code.

2. I really not sure how this would interaction with e.g. using `UserData::structured`: Having such an implementing for example based on `#[derive(Equivalence)]` should be fully sufficient to send structures over the wire even if they are not safely transmutable as the MPI implementation will only access valid data at specific offsets with correct type information. `FromAnyBytes` seems to be necessary only if one using an equivalence like
```
fn equivalent_datatype() -> Self::Out {
  UserDatatype::contiguous(mem::size_of::<Self>(), &u8::equivalent_datatype());
}
```
(which I personally use ATM for transmitting complex structs). My first current understanding would be that `FromAnyBytes` would not be an additional trait bound (in addition to `Equivalence`) but could rather provide a blanket impl of Equivalence like
```
unsafe impl<T> Equivalence for T where T: FromAnyBytes + ToBytes {
  type Out = UserDatatype;

  fn equivalent_datatype() -> Self::Out {
    UserDatatype::contiguous(mem::size_of::<Self>(), &u8::equivalent_datatype());
  }
}
```
But we would probably want people to opt in explicitly here, e.g. via a wrapper like `EquivalentToBytes(&mut T)`?

AndrewGaspar:
> That release seems to have happened in the mean time. I think the public dependency would actually be a better choice as interoperability w.r.t. these key traits could be automatic, e.g. other crates could implement this for their data types and we would automatically benefit. Ecosystem service effects and all that...

On further thought/discussion, it wasn't obvious, actually, that those traits are the best match. Essentially, we want a trait that only requires that the fields mapped by `Equivalence` can receive any bytes. Deriving `Equivalence` already requires the mapped fields to be valid objects in Rust.

> I would like to see MaybeUninit<M> included in the fold so that I can still receive arbitrary data types which are not safe to transmute

Good idea - done.

> I am really not sure how this would interact with e.g. using UserData::structured: Having such an implemention for example based on #[derive(Equivalence)] should be fully sufficient to send structures over the wire even if they are not safely transmutable

Yeah, you're absolutely right. I changed the name of the trait to `EquivalenceFromAnyBytes` to indicate that only those fields mapped by `Equivalence` require this attribute.

AndrewGaspar:
To address the issue with `bool`, I added a new type `mpi::Bool` that can receive any valid bit pattern, but has methods for unwrapping the value and checking validity. So you can still send `bool`, but only receive `mpi::Bool`. `mpi::Bool` is mapped as `MPI_C_BOOL`, just like `bool`, so it's valid as far as MPI is concerned to do this mapping.

README.md
```diff
@@ -174,19 +174,24 @@ comm.all_reduce_into(
...
 struct MyProgramOpts {
     name: [u8; 100],
     num_cycles: u32,
     material_properties: [f64; 20],
 }
 ```
 
+NOTE: `#[derive(EquivalenceFromAnyBytes)]` should be used in almost all circumstances. 
+This derive is required to receive into values of the designated type. Alternatively, you can wrap 
+your type in `MaybeUninit` to avoid the additional derive, but you'll need to verify the validity
```
adamreichold, line 192: I would suggest phrasing this slightly less operational, like "ensure" instead of "verify", as it might not be necessary to write any code for the validation, especially when running in more or less trusted environment as a HPC job often does.

build.rs
```diff
@@ -1,4 +1,11 @@
+use rustc_version::Version as RustcVersion;
+
 fn main() {
+    // Access to extern statics has to be marked unsafe after 1.13.0
+    if rustc_version::version().unwrap() >= RustcVersion::parse("1.46.0").unwrap() {
```
adamreichold, line 5: Personally, I think using the `autocfg` crate and testing for `#[track_caller]` using `probe_expression` might be nicer? (Also the comment seems to be copy&pasta?)

src/point_to_point.rs
```diff
@@ -1070,34 +1071,23 @@ pub trait MatchedReceiveVec {
...
-        Msg: Equivalence,
+        Msg: Equivalence + EquivalenceFromAnyBytes,
     {
         let (message, status) = self;
         let count = status
             .count(Msg::equivalent_datatype())
             .value_as()
             .expect("Message element count cannot be expressed as a usize.");
 
-        #[repr(transparent)]
```
adamreichold, line 1087: Thanks for taking the time to clean this up!

Cargo.toml
```diff
@@ -37,9 +37,12 @@ smallvec = "1.0.0"
 
 [build-dependencies]
 build-probe-mpi = { path = "build-probe-mpi", version = "0.1" }
+rustc_version = "0.2"
 
-# The following tests depend on specific features
+[dev-dependencies]
+safe-transmute = "0.11.0-rc.2"
```
adamreichold, line 43: Is this actually used anywhere now?

adamreichold:
@AndrewGaspar One thing I am wondering about here is usability. More specifically whether it is truly worth it to split the write and read sides of the equivalence traits. Is it really sufficiently common to have different sources for the sender and receiver processes as to significantly ease development of the write side of communication relation?

Personally, I think just folding the `EquivalenceFromAnyBytes` into the `Equivalence` trait would be simpler to understand and to explain without really making it more difficult to adhere to these requirements. Especially since the blanket implementation for `MaybeUninit` provides a generic escape hatch.

build.rs
adamreichold: On second thought, let's scratch that: This is not platform- or system-dependent but purely a compiler feature, so using `rustc_version` seems appropriate is surely much more efficient than `autocfg`.

README.md
AndrewGaspar: Well, I get what you're saying, but depending on the composite types (like `bool`, or perhaps an enum), there are UB considerations, as well. I can try to re-word it a bit, though.

AndrewGaspar:
I 100% get your concern and I'd like to eliminate it altogether. Here's where I'm stuck.

1. `Equivalence` designates an `MPI_Datatype` that safely maps the type. If MPI required type-checking, we'd be done here.
2. `EquivalenceFromAnyBytes` says: No matter what message you throw at me, my `Equivalence` implementation can safely (in Rust terms) map it into my representation
3. However, if you can't do that, that's OK: you can still wrap your type in `MaybeUninit`. As long as it implements `Equivalence`, it's still safe to receive into.

Here's the problem: if we roll  `EquivalenceFromAnyBytes` up into `Equivalence`, then we can't require `MaybeUninit<T>` in the cases where it's not safe to write arbitrary bytes into `T`.

Maybe if I'm understanding you correctly, the proposed alternative is to implement `Equivalence` on e.g. `MaybeUninit<bool>` instead? Though if that's the case, maybe we just only implement it on `mpi::Bool`.

README.md
adamreichold: > there are UB considerations

I would say any type confusion here would imply UB considerations. I think the question is rather whether this is checked for at runtime assuming it can happen in the given environment or whether it is assumed to not happen (and hence asserted with an `unsafe` block).

From my point of view, this poses the same problems for Rust as accessing structured data through memory maps. This is often handled via an `unsafe` block and suitable restrictions on the whole setup as well.

AndrewGaspar: Ah, gotcha, I see what you're saying now. Yeah, I think you're right.

adamreichold:
> Here's the problem: if we roll EquivalenceFromAnyBytes up into Equivalence, then we can't require MaybeUninit<T> in the cases where it's not safe to write arbitrary bytes into T.

I see.

> Maybe if I'm understanding you correctly, the proposed alternative is to implement Equivalence on e.g. MaybeUninit<bool> instead? Though if that's the case, maybe we just only implement it on mpi::Bool.

I think I would rather look at this from the other end: If `T: Equivalence`, the write API consumes `&T` whereas all read API produce `MaybeUninit<T>`. For types that are trivially transmutable, going from `MaybeUninit<T>` to `T` would be safe.

Of course, this would be a highly intrusive change to the API. There could also be a lot of friction if my model code contains e.g. a `Vec<T>` from which I send and into which I also want to receive, making it necessary to transmute `&mut Vec<T>` to `&mut Vec<MaybeUninit<T>>` even though that is always safe. (I am also not sure if `safe-transmute` does have the necessary functions to go from `T` to `MaybeUninit<T>` and back in safe way.)

But in the end, this would probably penalise users with types that would derive both `Equivalence` and `EquivalenceFromAnyBytes` now. So while I don't really feel comfortable about it, I agree that the currently proposal looks like the one with the lowest friction for the most common use case.

One last question on my mind is whether there is a trait in `safe-transmute` that we could use instead of `EquivalenceFromAnyBytes`? It does feel a bit like reinventing `TriviallyTransmutable`?

README.md
AndrewGaspar: I'll say something to the effect of:

"In HPC, the stakes of type confusion are relatively low, and it is recommended to just call `assume_init`. If you happen to be running MPI in other environments that accept untrusted user input, it is recommended to only work with types that implement `EquivalenceFromAnyBytes`.".

adamreichold:
> One last question on my mind is whether there is a trait in safe-transmute that we could use instead of EquivalenceFromAnyBytes? It does feel a bit like reinventing TriviallyTransmutable?

Ah, sorry, this is non-sense, as `EquivalenceFromAnyBytes` has a less strict contract, it means only what is touch by my `Equivalence` impl is `FromAnyBytes`, not the whole type. Sorry for the confusion!

AndrewGaspar:
> Ah, sorry, this is non-sense, as EquivalenceFromAnyBytes has a less strict contract, it means only what is touch by my Equivalence impl is FromAnyBytes, not the whole type. Sorry for the confusion!

Was just about to say this - exactly. That's why I opted not to use it.

AndrewGaspar:
Take one last look?

adamreichold:
> Take one last look?

One more idea I had concerning usability: Maybe we should make `EquivalenceFromAnyBytes` a sub trait of `Equivalence` as it really only makes sense for a type that actually implements `Equivalence` in the first place?

This would also suggest to make the derive macro for `EquivalenceFromAnyBytes` derive both impls so that I can just write
```
#[derive(EquivalenceFromAnyBytes)]
```
instead of
```
#[derive(Equivalence, EquivalenceFromAnyBytes)]
```
One downside I see would be being unable to manually implement `Equivalence` but derive `EquivalenceFromAnyBytes`, but then again manually implementing `EquivalenceFromAnyBytes` on top of manually implementing `Equivalence` might be the better trade-off if the common case is simplified?

build.rs
```diff
@@ -1,4 +1,11 @@
+use rustc_version::Version as RustcVersion;
+
 fn main() {
+    // Access to extern statics has to be marked unsafe after 1.13.0
```
adamreichold, line 4: This comment still does not match the code. :-)

src/checked_bool.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    /// Coerce `Bool` to `bool`.
+    #[cfg_attr(track_caller_supported, track_caller)]
+    pub fn unwrap(&self) -> bool {
+        self.valid().unwrap()
+    }
+
+    /// Unwraps `Bool` if valid, else returns `default`.
```
adamreichold, line 60: I am not sure whether this should have `track_caller` attached as well? But more generally, I guess we should not add this because if this, why not `unwrap_or_else` and any other helper method on `Result`? The pattern `self.valid().foo(..)` seems simple enough IMHO.

src/topology/mod.rs
```diff
@@ -705,22 +705,24 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
     /// # Standard Sections
     ///
     /// 4.2, see MPI_Unpack
-    unsafe fn unpack_into<Buf>(&self, inbuf: &[u8], outbuf: &mut Buf, position: Count) -> Count
```
adamreichold, line 708: Just to make sure I understand this correctly: This has become safe because `BufferMut` now requires `EquivalenceFromAnyBytes`?

I guess there also were quite a few functions that should have been unsafe for the same reasons, but were not which is why this is the only place that has changed from being unsafe to safe due to the `BufferMut` change?

src/checked_bool.rs
AndrewGaspar: Yeah, you're probably right. I'll remove the othe member functions.

src/topology/mod.rs
AndrewGaspar: Yeah, that's exactly right. The addition of `unpack_into` is actually what started discussion about safe transmutability. At first we panicked and thought maybe we needed to make every single function that deals with an MPI buffer unsafe, but then decided that it was probably safe to assume the bounds of bad outcomes with mismatched typemaps was pretty low.

AndrewGaspar:
> > Take one last look?
> 
> One more idea I had concerning usability: Maybe we should make `EquivalenceFromAnyBytes` a sub trait of `Equivalence` as it really only makes sense for a type that actually implements `Equivalence` in the first place?
> 
> This would also suggest to make the derive macro for `EquivalenceFromAnyBytes` derive both impls so that I can just write
> 
> ```
> #[derive(EquivalenceFromAnyBytes)]
> ```
> 
> instead of
> 
> ```
> #[derive(Equivalence, EquivalenceFromAnyBytes)]
> ```
> 
> One downside I see would be being unable to manually implement `Equivalence` but derive `EquivalenceFromAnyBytes`, but then again manually implementing `EquivalenceFromAnyBytes` on top of manually implementing `Equivalence` might be the better trade-off if the common case is simplified?

Yeah, I had considered doing exactly what you're describing, but decided not to for the exact downsides you described. I also wasn't sure if automatically implementing multiple traits for a single derive was considered good practice.

I don't know, I think I'm still leaning towards keeping them as separate derives because I think that's the maximally flexible option.

adamreichold:
> I don't know, I think I'm still leaning towards keeping them as separate derives because I think that's the maximally flexible option.

Alright. I'd say we keep it as-is then and collect some experience actually using it. We are at 0.6 instead of 1.6 after all.

src/checked_bool.rs
adamreichold: Gnarrr, it seems they are still used in some unit test.

AndrewGaspar: 🤦‍♂️ I always forget to run `cargo test` even though I wrote the dang tests. At some point I'll figure out how to get all the "example" tests into the appropriate place so it's all just `cargo test`.

src/transmute.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use std::mem::MaybeUninit;
+
+use crate::datatype::Equivalence;
+
+pub(crate) mod traits {
+    pub use super::EquivalenceFromAnyBytes;
+}
+
+/// Any type whose `Equivalence` implementation maps only fields that can be composed of an
```
adamreichold, line 14: I am sorry that it never seems to stop, but one more question came to my mind on this: If one has a struct with three fields, say one is a `bool`, and an `Equivalence` that maps only the other two which are `EquivalenceFromAnyBytes`, then this could implement `EquivalenceFromAnyBytes` as it fulfills what is written here, but that would also mean that I can receive a partially formed instance via MPI (as the bool field is not mapped  and hence not initialised). On the other hand, wrapping it received value in `MaybeUninit` makes this is workable `Equivalence` implementation, one just has to initialise the `bool` field manually after receiving and before calling `assume_init`.

Hence I think we should reword this in a more operational way, i.e. what we want is that the `Equivalence` impl of the type in question produces a fully initialised value from any bit pattern received via MPI.

I think this would also suggest adjusting the naming a bit to something like `Full...`, `Complete...`, `Safe...` or UnconditionalEquivalence`?

I also wonder about the derive macro for `EquivalenceFromAnyBytes`: The above problem even exists if the unmapped field is say `u32` as while this `TriviallyTransmutable`, it would still be unmapped and hence uninitialised when receiving the type via MPI (it is not insta-UB, just UB I guess). So just checking the type of all fields is not sufficient if we do not know if all fields are actually mapped.

For me, this again suggests making this a super trait of `Equivalence` and also making the derive macro for it derive both traits because this would make sure that all fields are mapped and the check if those field types are themselves `EquivalenceFromAnyBytes` sound.

AndrewGaspar: > On the other hand, wrapping it received value in MaybeUninit makes this is workable Equivalence implementation, one just has to initialise the bool field manually after receiving and before calling assume_init.

You don't even really need `MaybeUninit`. In pure safe code, you can just initialize the receive buffer like normal before hand and there won't be any undefined behavior.

> Hence I think we should reword this in a more operational way, i.e. what we want is that the Equivalence impl of the type in question produces a fully initialised value from any bit pattern received via MPI.

I don't think that's necessary. As long as you're correctly initializing your receive buffer, including unmapped fields, then there is no issue.

adamreichold: > As long as you're correctly initializing your receive buffer, including unmapped fields, then there is no issue.

This only applies to methods where I pass an already initialised value in, which is then "merely updated". But things like `Source::receive` and `Source::receive_vec` produce their return types `Msg: Equivalence` using only the data passed in via MPI.

These functions would also need updated trait bounds if I understand things correctly as adding `BufferMut: EquivalenceFromAnyBytes` does not affect them as they work using `with_uninitialized2` internally.

AndrewGaspar: Crap, good point. But I don't think we can guard against too-small messages, though, even if we made the trait stricter. Let me test this on my system to see...

In the cases what if we just required either implementing `Default` for your type, or otherwise wrapping it in `MaybeUninit`?

AndrewGaspar: > But I don't think we can guard against too-small messages, though, even if we made the trait stricter. Let me test this on my system to see...

Nvm, OpenMPI will crash if the message is too small to fill the receive buffer. Let me see with mpich, too.

adamreichold: > In the cases what if we just required either implementing Default for your type, or otherwise wrapping it in MaybeUninit?

Requiring `Default` would certainly suffice but it makes for a rather limited API and I think we should avoid it.

Returning `MaybeUninit<Msg>` would also work, but I think that keeping that route viable via `impl FullEquivalence for MaybeUninit<Msg> where Msg: Equivalence` would be a nicer way to include this.

> But I don't think we can guard against too-small messages, though, even if we made the trait stricter. Let me test this on my system to see...

I would assume that the MPI implementation has to make sure that the received data can be successfully used with the given data type, whether that contains "padding" (including unmapped fields) or not, as we never get direct access to the raw bytes, only the results of mapping them through or MPI data type.

AndrewGaspar: So MPICH appears to *rubbing temples* completely ignore the size of the input buffer in `MPI_Unpack` and just attempts to read the input buffer as if it was at least the size of the output buffer.

Test code:
```rust
use mpi::traits::*;

fn main() {
    let universe = mpi::initialize().unwrap();

    let comm = universe.world();

    #[derive(Equivalence, EquivalenceFromAnyBytes)]
    struct MySend {
        a: u32,
    }

    #[derive(Equivalence, EquivalenceFromAnyBytes, Debug)]
    struct MyRecv {
        a: u32,
        b: u32,
        nums: [u32; 256],
    }

    let mut recv = MyRecv {
        a: 0x0,
        b: 0x10,
        nums: [0xFFFFFFFF; 256],
    };
    comm.unpack_into(comm.pack(&MySend { a: 0x5 }).as_slice(), &mut recv, 0);

    dbg!(recv);
}
```

AndrewGaspar: Getting a bunch of garbage in the `nums` field.

adamreichold: The more I think about this, I find the approach that extends `Equivalence` with a marker trait saying "whatever bit pattern is mapped through my `Equivalence` impl, it will produce a valid instance of my type" most useful.

I also think the "complicated" scenario where I map only some types through MPI but those that I do map are `TriviallyTransmutable` is better served my `MaybeUninit` if only to make these complications more visible in user code.

adamreichold: > So MPICH appears to *rubbing temples* completely ignore the size of the input buffer in MPI_Unpack and just attempts to read the input buffer as if it was at least the size of the output buffer.

I would be inclined to call this a bug in MPICH as we explicitly pass `inbuf.count()` as the input size which it should not ignore lest it reads uninitialised memory, whether we call it from Rust code or C/C++ code. (I guess we could try to validate the size using `MPI_Pack_size` but that would be redundant work for e.g. OpenMPI.)

> Getting a bunch of garbage in the nums field.

Out of curiosity, what is the position returned in your test case?

AndrewGaspar: `std::mem::size_of::<MyRecv>()`. 🤦‍♂️

AndrewGaspar: Which is 1032

AndrewGaspar: > I would be inclined to call this a bug in MPICH as we explicitly pass inbuf.count() as the input size which it should not ignore lest it reads uninitialised memory, whether we call it from Rust code or C/C++ code. (I guess we could try to validate the size using MPI_Pack_size but that would be redundant work for e.g. OpenMPI.)

Filed a couple issues to fix this after 0.6.0. I say we just leave the MPICH-specific unsoundness in there for now. See #102 #103

AndrewGaspar: OK, I expanded the semantics of `EquivalenceFromAnyBytes` to require that the the type can be fully instantiated from an arbitrary, yet properly sized MPI message. If you feel strongly about the name, I can still change it...

To that end, I fixed up a bunch of routines that directly instantiate the receive buffer to ensure that the received message is the correct size.

> For me, this again suggests making this a super trait of Equivalence and also making the derive macro for it derive both traits because this would make sure that all fields are mapped and the check if those field types are themselves EquivalenceFromAnyBytes sound.

Yeah, I think at this point we have to... Well, there is an alternative option - if the user derives `EquivalenceFromAnyBytes`, we could require they also derive `Equivalence`. I think I'll try that.

AndrewGaspar: > Well, there is an alternative option - if the user derives EquivalenceFromAnyBytes, we could require they also derive Equivalence. I think I'll try that.

nvm, you can't get the other derives. I'll just have `EquivalenceFromAnyBytes` also derive `Equivalence`.

AndrewGaspar: So now the derives are just `Equivalence` and `EquivalenceUnsafe`. The former impls both traits, and the latter just impls `Equivalence`.

adamreichold: > If you feel strongly about the name, I can still change it...

I do not feel strongly about it all. I do wonder whether the name of the derive macro not matching up with the traits could cause confusion though?

I also did not read through the updated code yet...

src/point_to_point.rs
```diff
@@ -146,25 +146,29 @@ pub unsafe trait Source: AsCommunicator {
...
-                )
+                );
             });
             let status = Status(status);
-            if status.count(Msg::equivalent_datatype()) == 0 {
-                panic!("Received an empty message.");
-            }
-            (msg, status)
+
+            let count = status.count(&datatype).expect(
```
adamreichold, line 167: Just for readability, couldn't you move the checking of `status` and `count` (meaning the panicking essentially) inside the closure passed to `with_uninitialized2` and thereby avoid manually juggling the `MaybeUninit`?

src/transmute.rs
adamreichold: One thing I would still find nice is if `EquivalenceFromAnyBytes: Equivalence`: For one thing the first trait really adds a guarantee on the impl of the second the trait, so implementing `EquivalenceFromAnyBytes` for a type that does not implement `Equivalence` really does not make sense. And it would also simplify the trait bounds used in our code base somewhat as `Equivalence + EquivalenceFromAnyBytes` would collapse to `EquivalenceFromAnyBytes`.

AndrewGaspar: Yeah, I'll change the the trait names to match.

AndrewGaspar: Just realized that I can't impl `EquivalenceUnsafe` for `MaybeUninit<T>` where `T: EquivalenceUnsafe`. It would make it so you could send a `MaybeUninit<T>`. I need to basically only enable receiving into `MaybeUninit<T>`...

src/point_to_point.rs
adamreichold: I think I had a think'o here: `status` is `MaybeUninit` itself and we need its `assume_init` to be called before making the check. That sad, I wonder if `with_uninitialized` helps here at all as it somewhat obscures where `assume_init` is called? Maybe we need a receive helper to encapsulate this repeating two step check?

src/transmute.rs
adamreichold: > Just realized that I can't impl `EquivalenceUnsafe` for `MaybeUninit<T>` where `T: EquivalenceUnsafe`. It would make it so you could send a `MaybeUninit<T>`. I need to basically only enable receiving into `MaybeUninit<T>`...

I understand it looks suspicious but I don't think there is anything inherently wrong with sending `MaybeUninit<T>`? Of course, the receiver should not call `assume_init` if the value sent was not fully initialised, but I don't think MPI would care.

adamreichold: >  I need to basically only enable receiving into MaybeUninit<T>...

I think the only way to really achieve this to provide separate API, e.g. `receive_uninit` which returns `MaybeUninit<Msg>` instead of `Msg` but has the trait bound `Msg: Equivalence` instead of `Msg: EquivalenceFromAnyBytes`? But that seems like a hefty complication of the API...

adamreichold: > I understand it looks suspicious but I don't think there is anything inherently wrong with sending MaybeUninit<T>? 

Or put differently, sending `MaybeUninit<T>` but receiving `T` is "just" type confusion similar to sending `u8` but receiving `bool`.

AndrewGaspar: > I understand it looks suspicious but I don't think there is anything inherently wrong with sending MaybeUninit<T>? Of course, the receiver should not call assume_init if the value sent was not fully initialised, but I don't think MPI would care.

That's not right - it would allow you to read uninitialized memory without an unsafe block.

```
    let mut data = [0xFFu8; 1024];
    comm.unpack_into(
        comm.pack(&[MaybeUninit::<u8>::uninit(); 1024][..])
            .as_slice(),
        &mut data[..],
        0,
    );

    println!("{:?}", data);
```

adamreichold: Yes, indeed. When the data really is uninitialised it read without an `unsafe` block.

Hhhmmm, so `EquivalenceFromAnyBytes` cannot be a sub trait of `Equivalence` because we want `MaybeUninit<T>: EquivalenceFromAnyBytes` if `T: Equivalence` but we do _not_ want `MaybeUninit<T>: Equivalence`. But even this does not really help us as we still we need to have separate `receive_uninit<Msg>() -> MaybeUninit<Msg>` API to express that `Msg: Equivalence` and `MaybeUninit<Msg>: EquivalenceFromAnyBytes`, but this would also mean that we do not need MaybeUninit<T>: EquivalenceFromAnyBytes` at all as this is implicit in the `receive_unit` signature, i.e. separate entry points for uninitialised messages make the blanket impl redundant.

Should we split this into three traits `Equivalence`, `SafeToSend` and `SafeToReceive` where `MaybeUninit<T>` would be `SafeToReceive` but not `SafeToSend`? This does not really sound better than the separate API entry points, does it? Especially since the API entry points for initialised messages could be forwarding implementations like
```
fn receive<Msg>(&self) -> Msg where Msg: EquivalenceFromAnyBytes { unsafe { self.receive_uninit().assume_init() } }
```

adamreichold: (I would still argue to make `EquivalenceFromAnyBytes` a sub trait of `Equivalence` even if we need handle `MaybeUninit` separately.)

adamreichold: ((Just so this will never end (really sorry for this being so complicated): I think at this point `EquivalenceFromAnyBytes` really is `Equivalence + TriviallyTransmutable` so we could reuse the trait from `safe-transmute`.))

adamreichold: @AndrewGaspar I tried making progress into any of the above directions and laziness pushed me towards introducing an additional trait which matches the two situations where are interested in instead of increasing the API surface, i.e.
```rust
pub unsafe trait SafeToReceive {
    type Inner: Equivalence;
}

unsafe impl<T> SafeToReceive for T
where
    T: Equivalence + EquivalenceFromAnyBytes,
{
    type Inner = T;
}

unsafe impl<T> SafeToReceive for MaybeUninit<T>
where
    T: Equivalence,
{
    type Inner = T;
}
```
c.f. https://github.com/adamreichold/rsmpi/commit/2e228673db0ebb5398ee96d68c789894f5887f83

Do you think this would be a viable approach to move this change forward?

(I still think we might want to go for a sub trait or even `TriviallyTransmutable`, but by now, I think it would be wiser to postpone any such changes to separate issues/pull requests.)

AndrewGaspar:
@adamreichold Sorry I haven't checked in on this in awhile - I've been going through a long interview process with a few different companies, and haven't had much free time. Thankfully, I'm finally done with all that, and I may even get to write Rust full time in my new job! 🤞 I'm going to try getting this done when I leave my current job next week - I have four weeks off before I start my next job.

src/datatype.rs
```diff
@@ -197,21 +223,63 @@ equivalent_system_datatype!(i8, ffi::RSMPI_INT8_T);
...
+unsafe impl<T> Equivalence for MaybeUninit<T>
+where
+    T: Equivalence,
+{
+    type Out = T::Out;
+
+    fn equivalent_datatype() -> Self::Out {
+        T::equivalent_datatype()
+    }
+}
```
jedbrown, line 282: I think this is a problem since it's UB to send from an uninitialized buffer. I think MaybeUninit should be used with an interface that returns a reference to the message as received (e.g., as a slice of the `MaybeUninit` receive buffer), as discussed here. https://github.com/rsmpi/rsmpi/pull/128#issuecomment-1280034891

examples/complex_datatype.rs
```diff
@@ -1,22 +1,34 @@
...
+#[repr(C)]
+struct TupleType(
+    [f32; 2],
+    u8,
+    u8,  // unused
+    u16, // unused
+);
+
+// All subfields are `FromAnyBytes`.
+unsafe impl EquivalenceFromAnyBytes for TupleType {}
```
jedbrown, line 24: I believe we need the stronger statement that there is no padding. For example, consider where bytemuck requires [`NoUninit`](https://docs.rs/bytemuck/latest/bytemuck/trait.NoUninit.html) versus `Pod`. I think for a symmetric interface, we basically need `Pod` and perhaps we should have rsmpi recognize bytemuck traits rather than making our own traits (which have fuzzy semantics at present).

examples/complex_datatype.rs
```diff
@@ -48,25 +60,39 @@ unsafe impl Equivalence for ComplexDatatype {
     }
 }
 
+// All subfields are `FromAnyBytes`.
```
jedbrown, line 63: I think we also need that there are no padding/uninit parts, which is a hard thing to test visually, but the bytemuck derive macro does check.

jedbrown (commented):
@adamreichold @AndrewGaspar :wave: What do you think should be done with this PR? I'll admit to being kinda worried about what it sets out to do, given that MPI says mismatched types are invalid (even though implementations don't check). Based on @jtronge's work on our paper (https://dl.acm.org/doi/fullHtml/10.1145/3615318.3615328), I'd rather create a safe way to check for type matching, such as creating lightweight typed communicators where type matching can be validated by nonblocking collectives or metadata piggy-backed on the first message (or possibly MPI extensions, which the paper show can have small overhead). Perhaps these efforts are complementary, but I'm still not sure how to think about this PR. (I'm not asking for any development time, just your opinion about what effort would be fruitful.)

PR #83, opened 2019-12-22
Fix equivalent_datatype() in examples/structured.rs

hamza-tam:
The example provided does not compile, because of the missing `count` parameter and the conflicting type of the `types` parameter.


AndrewGaspar:
Hi, this example is correct with the latest API changes in 0.6, we just haven't published a release since the change was made.

I'm closing this issue. If you have concerns with the API changes in the next version, please open a separate issue.

PR #88, opened 2020-03-08
Implemented MPI_Waitany

finoalessiopolimi:
I needed the MPI_Waitany for a personal project so I decided to implement it as a function in the Request module. This is the first time i work with ffi and unsafe code so I tried to keep things as simple as possible.
I'm not sure if using a Vec of Request as input parameter and removing the completed Request from it is the best way to go but it was the simplest I could find.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -88,46 +88,38 @@ impl<'a, S: Scope<'a>> Drop for Request<'a, S> {
 ///
 /// The completed request is removed from the vector of requests.
 ///
-/// If no Request is active the index returned is mpi_sys::MPI_UNDEFINED.
+/// If no Request is active None is returned.
 ///
 /// # Examples
 ///
 /// See `examples/wait_any.rs`
-pub fn wait_any<'a, S: Scope<'a>>(requests: &mut Vec<Request<'a, S>>) -> (i32, Status) {
-    let mut mpi_requests: Vec<MPI_Request> = Vec::with_capacity(requests.len());
-    let mut scopes: Vec<S> = Vec::with_capacity(requests.len());
+pub fn wait_any<'a, S: Scope<'a>>(requests: &mut Vec<Request<'a, S>>) -> Option<(usize, Status)> {
+    let mut mpi_requests: Vec<_> = requests.iter().map(|r| r.as_raw()).collect();
+    let mut index: i32 = mpi_sys::MPI_UNDEFINED;
+    let size: i32 = mpi_requests
+        .len()
+        .try_into()
+        .expect("Error while casting usize to i32");
+    let status;
     unsafe {
-        let mut index: i32 = mpi_sys::MPI_UNDEFINED;
-        while let Some(r) = requests.pop() {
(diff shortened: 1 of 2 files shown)


src/request.rs
```diff
@@ -82,6 +83,54 @@ impl<'a, S: Scope<'a>> Drop for Request<'a, S> {
...
+///
+/// # Examples
+///
+/// See `examples/wait_any.rs`
+pub fn wait_any<'a, S: Scope<'a>>(requests: &mut Vec<Request<'a, S>>) -> (i32, Status) {
+    let mut mpi_requests: Vec<MPI_Request> = Vec::with_capacity(requests.len());
+    let mut scopes: Vec<S> = Vec::with_capacity(requests.len());
+    unsafe {
+        let mut index: i32 = mpi_sys::MPI_UNDEFINED;
+        while let Some(r) = requests.pop() {
```
AndrewGaspar: Rather than draining the `requests` object and then re-filling it, could you instead just build the `mpi_requests` objects using `as_raw`?

e.g. 
```rust
let mut mpi_requests: Vec<_> = requests.iter().map(|r| r.as_raw()).collect();
```

Then after `MPI_Waitany` succeeds, just do `requests.remove(index);`. You'll need to make sure that there's no panic between `MPI_Waitany` and the `remove` call.

I think this approach has two benefits:
1) Eliminates the allocation of the `scopes` vector
2) Fewer moves in the `requests` vector.

src/request.rs
```diff
@@ -82,6 +83,54 @@ impl<'a, S: Scope<'a>> Drop for Request<'a, S> {
     }
 }
 
+/// Wait for the completion of one of the requests in the vector,
+/// returns the index of the request completed and the status of the request.
+///
+/// The completed request is removed from the vector of requests.
+///
+/// If no Request is active the index returned is mpi_sys::MPI_UNDEFINED.
```
AndrewGaspar: return an `Option<(i32, Status)>` instead - prefer Rust idioms when there's a clear mapping.

src/request.rs
AndrewGaspar: Oh, you're also not reversing the list and then don't have to deal with fixing up the index.

src/request.rs
```diff
@@ -82,6 +83,54 @@ impl<'a, S: Scope<'a>> Drop for Request<'a, S> {
...
+///
+/// # Examples
+///
+/// See `examples/wait_any.rs`
+pub fn wait_any<'a, S: Scope<'a>>(requests: &mut Vec<Request<'a, S>>) -> (i32, Status) {
```
AndrewGaspar: Let's actually make this a `usize` - we're already checking that it's convertible anyway.

AndrewGaspar:
By the way, stay strong. America stands with Italy!

finoalessiopolimi:
Thank you Andrew, I hope the situation there in America don't get as bad as here.

I've fixed the problems you have spotted in the implementation.

PR #89, opened 2020-08-09
Add MPI_Comm_split_type MPI_COMM_TYPE_SHARED function.

zolkko:
In a project of mine, I am using this function to group processes based on their physical compute node. And later a group's root can control and share resources with processes in a group.

Only MPI_COMM_TYPE_SHARED split type is covered, because, If as I understand, it is the only obligatory split type described in the standard.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -474,6 +474,18 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


AndrewGaspar:
Hi, I had some feedback, but I figured it was easier to just make the changes myself. Check out #90 - if the changes are to your liking, please close this PR. Thanks!

PR #90, opened 2020-08-10
Add MPI_Comm_split_type MPI_COMM_TYPE_SHARED function [AMENDED]

AndrewGaspar:
I addressed some concerns I had with #89. I moved the routine to a different location in the file (roughly sorted by declaration in the standard). I also documented where in the standard `MPI_Comm_split_type` is described. Additionally, I returned just `UserCommunicator` rather than `Option<UserCommunicator>` - at least by my reading of the standard, if `MPI_COMM_TYPE_SHARED` is specified, a new communicator is always created. `MPI_COMM_NULL` seems to only be returned if you specified `MPI_UNDEFINED`. Finally, I created `RSMPI_COMM_TYPE_SHARED` - some implementations declare `MPI_COMM_TYPE_SHARED` using `#define` which isn't picked up by bindgen.

@zolkko - please take a look.


AndrewGaspar:
Sorry, I'm running into issues with the CI. Haven't gotten the chance to look into why it's failing yet. Seems to be an infrastructure issue.

AndrewGaspar:
It seems to be failing in master in the same way, so I'm just going to merge and fix in a separate PR.

PR #91, opened 2020-08-18
Switch to GitHub Actions

AndrewGaspar:
Travis CI has been throwing up some bizarre timeouts, so this is an attempt to switch to GitHub Actions, which I've generally had good experiences with


AndrewGaspar:
Add a "clippy" job, too. See: https://github.com/rsmpi/rsmpi/pull/101#pullrequestreview-533343821

jedbrown:
Alright, this passes with I think decent coverage. Thanks @AndrewGaspar for the groundwork, especially the Windows install.

I'm going to merge this now because there's presently no CI on `main` -- we can modify this in a future PR if there are comments.

Issue #57, opened 2018-11-08
Cannot build 32-bit MS-MPI application on 64-bit Windows


AndrewGaspar:
NOTE: This is just a note for my own sake. Low priority.

We pick the link library location to use for MS-MPI based on target_arch:
```rust
    let lib_env = if cfg!(target_arch = "i686") {
        "MSMPI_LIB32"
    } else if cfg!(target_arch = "x86_64") {
        "MSMPI_LIB64"
    } else {
        panic!("rsmpi does not support your windows architecture!");
    };
```

Of course, build scripts are always compiled for the host architecture. We need to use the "TARGET" [environment variable](https://doc.rust-lang.org/cargo/reference/environment-variables.html) to get the triple for the actual target.

Issue #44, opened 2018-03-23
Build error because of different clang-sys/bindgen requirements


5jqk2ymwqd:
When I tried to build a project depending on this crate I get the following error:
> error: multiple packages link to native library `clang`, but a native library can be linked only once
> 
> package `clang-sys v0.21.2`
>     ... which is depended on by `bindgen v0.31.3`
>     ... which is depended on by `libffi-sys v0.6.0`
>     ... which is depended on by `libffi v0.6.3`
>     ... which is depended on by `mpi v0.5.2`
>     ... which is depended on by `project-using-mpi v0.1.0 (file:///path/to/project-folder)`
> links to native library `clang`
> 
> package `clang-sys v0.22.0`
>     ... which is depended on by `bindgen v0.33.2`
>     ... which is depended on by `mpi-sys v0.1.1`
>     ... which is depended on by `mpi v0.5.2`
>     ... which is depended on by `project-using-mpi v0.1.0 (file:///path/to/project-folder)`
> also links to native library `clang`

The project has only mpi as dependency (Cargo.toml):
> [dependencies]
> mpi = "*"

Maybe `mpi-sys` could depend on `bindgen v0.31.3` (libffi-sys depends [explicitly](https://github.com/tov/libffi-sys-rs/blob/master/Cargo.toml) on bindgen = "0.31.3"), too. Or do you think it is better to wait for `libffi-sys` to update to the current `bindgen`?

Thanks a bunch for this crate, I am excited to try it.

bsteinb:
Hm, that is unfortunate, I will tie down the version of `bindgen` that `mpi-sys` uses and publish a new version to crates.io.

However, I suspect that `cargo` is being too cautious in this situation. `bindgen` and through it `clang-sys` is a build dependency of both `mpi-sys` and `libffi-sys`, so it will be linked to their `build` scripts, but not to the resulting libraries, so there will not be a mix of different versions of `clang-sys` or the native `libclang` in the resulting `mpi` library. I will open an issue for `cargo` and see what the developers think about this.

Thank you for the report and the encouraging words.

bsteinb:
All right, I have just released `mpi-0.5.3` which should fix the issue.

5jqk2ymwqd:
Thank you very much.

Eh2406:
Oddly I can not reproduce, I built a new project and added 
```
[dependencies]
mpi = "=0.5.2"
```
[...]

PR #96, opened 2020-10-25
Update libffi, bindgen, and restore user-operations as a default feature

AndrewGaspar:
As of the "0.9.0" release of the libffi crate, Windows and Mac are now supported. Additionally, in "1.0.0", bindgen is no longer used by libffi, which means we can now update our version of the dependency.

In order to avoid surprising people with the "0.6" release, this makes "user-operations" a default feature again since it now has wider support.

Improved behavior in the "0.55" release of bindgen gets rid of the weird hacks that had to be done for MS-MPI.

cc @adamreichold @bsteinb

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,15 +1,7 @@
 use std::env::var;
 
 fn main() {
-    let is_msmpi = if let Ok(library) = build_probe_mpi::probe() {
-        if library.version == "MS-MPI" {
-            true
-        } else {
-            false
-        }
-    } else {
-        false
-    };
+    let is_msmpi = build_probe_mpi::probe().map_or(false, |lib| lib.version == "MS-MPI");
 
     if is_msmpi {
         println!("cargo:rustc-cfg=msmpi");
--- a/build.rs
+++ b/build.rs
@@ -1,18 +1,7 @@
-use std::env::var;
-
 fn main() {
     let is_msmpi = build_probe_mpi::probe().map_or(false, |lib| lib.version == "MS-MPI");
 
     if is_msmpi {
         println!("cargo:rustc-cfg=msmpi");
-
-        if var("CARGO_FEATURE_USER_OPERATIONS").is_ok()
-            && var("CARGO_CFG_TARGET_ARCH") == Ok("x86".to_string())
-        {
-            panic!(
-                "Feature 'user-operations' is not supported for MS-MPI on 32-bit Windows. \
-            See: https://github.com/rsmpi/rsmpi/issues/97"
-            )
-        }
     }
 }
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,8 @@
 fn main() {
-    let is_msmpi = build_probe_mpi::probe().map_or(false, |lib| lib.version == "MS-MPI");
+    let is_msmpi = match build_probe_mpi::probe() {
+        Ok(lib) => lib.version == "MS-MPI",
+        _ => false,
+    };
 
     if is_msmpi {
         println!("cargo:rustc-cfg=msmpi");
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,26 @@
+use std::env::var;
+
 fn main() {
-    if cfg!(windows) {
-        // Adds a cfg to identify MS-MPI. This should perhaps be a more robust check in the future.
+    let is_msmpi = if let Ok(library) = build_probe_mpi::probe() {
+        if library.version == "MS-MPI" {
+            true
+        } else {
+            false
+        }
+    } else {
+        false
+    };
+
+    if is_msmpi {
         println!("cargo:rustc-cfg=msmpi");
+
+        if var("CARGO_FEATURE_USER_OPERATIONS").is_ok()
+            && var("CARGO_CFG_TARGET_ARCH") == Ok("x86".to_string())
+        {
+            panic!(
+                "Feature 'user-operations' is not supported for MS-MPI on 32-bit Windows. \
+            See: https://github.com/rsmpi/rsmpi/issues/97"
+            )
+        }
     }
 }
--- a/mpi-sys/build.rs
+++ b/mpi-sys/build.rs
@@ -21,25 +21,25 @@ fn main() {
         }
     };
 
-    // Use `mpicc` wrapper on Unix rather than the system C compiler.
+    let mut builder = cc::Build::new();
+    builder.file("src/rsmpi.c");
+
+    let compiler = builder.try_get_compiler();
+
     if cfg!(windows) {
-        let mut builder = cc::Build::new();
-
-        builder.file("src/rsmpi.c");
-
         for inc in &lib.include_paths {
             builder.include(inc);
         }
 
-        builder.compile("librsmpi.a");
-
         // Adds a cfg to identify MS-MPI
         println!("cargo:rustc-cfg=msmpi");
     } else {
+        // Use `mpicc` wrapper on Unix rather than the system C compiler.
         env::set_var("CC", "mpicc");
-        // Build the `rsmpi` C shim library.
-        cc::Build::new().file("src/rsmpi.c").compile("librsmpi.a");
     }
+
+    // Build the `rsmpi` C shim library.
+    builder.compile("rsmpi");
 
     // Let `rustc` know about the library search directories.
     for dir in &lib.lib_paths {
@@ -53,6 +53,22 @@ fn main() {
     // Let `bindgen` know about header search directories.
     for dir in &lib.include_paths {
         builder = builder.clang_arg(format!("-I{}", dir.display()));
+    }
+
+    // Get the same system includes as used to build the "rsmpi" lib. This block only really does
+    // anything when targeting msvc.
+    if let Ok(compiler) = compiler {
+        let include_env = compiler
+            .env()
+            .iter()
+            .filter(|(key, _)| key == "INCLUDE")
+            .next();
+        if let Some((_, include_paths)) = include_env {
+            if let Some(include_paths) = include_paths.to_str() {
+                // Add include paths via -I
+                builder = builder.clang_args(include_paths.split(";").map(|i| format!("-I{}", i)));
+            }
+        }
     }
 
     // Generate Rust bindings for the MPI C API.
--- a/mpi-sys/build.rs
+++ b/mpi-sys/build.rs
@@ -24,8 +24,6 @@ fn main() {
     let mut builder = cc::Build::new();
     builder.file("src/rsmpi.c");
 
-    let compiler = builder.try_get_compiler();
-
     if cfg!(windows) {
         for inc in &lib.include_paths {
             builder.include(inc);
@@ -35,8 +33,10 @@ fn main() {
         println!("cargo:rustc-cfg=msmpi");
     } else {
         // Use `mpicc` wrapper on Unix rather than the system C compiler.
-        env::set_var("CC", "mpicc");
+        builder.compiler("mpicc");
     }
+
+    let compiler = builder.try_get_compiler();
 
     // Build the `rsmpi` C shim library.
     builder.compile("rsmpi");
--- a/mpi-sys/build.rs
+++ b/mpi-sys/build.rs
@@ -58,11 +58,7 @@ fn main() {
     // Get the same system includes as used to build the "rsmpi" lib. This block only really does
     // anything when targeting msvc.
     if let Ok(compiler) = compiler {
-        let include_env = compiler
-            .env()
-            .iter()
-            .filter(|(key, _)| key == "INCLUDE")
-            .next();
+        let include_env = compiler.env().iter().find(|(key, _)| key == "INCLUDE");
         if let Some((_, include_paths)) = include_env {
             if let Some(include_paths) = include_paths.to_str() {
                 // Add include paths via -I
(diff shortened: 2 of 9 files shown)


adamreichold:
Just to make sure: We still need `cargo:rustc-cfg=msmpi` in `build.rs` after this change?

(As a tangentially related question, do we need `cargo:rustc-cfg=extern_statics_are_unsafe` still? Do we have no transitive dependency that implies an MSRV of 1.13? Could just always use `unsafe` and ignore `unused_unsafe` warnings on older compilers?)

AndrewGaspar:
> Just to make sure: We still need cargo:rustc-cfg=msmpi in build.rs after this change?

Yes, it disables a handful of APIs that are not available with MS-MPI.

> do we need cargo:rustc-cfg=extern_statics_are_unsafe still?

Probably not - Rust 2018 is required now, after all.

adamreichold:
> Probably not - Rust 2018 is required now, after all.

Would be nice to remove this then, but probably not in this PR I guess as it is rather unrelated to the change at hand.

AndrewGaspar:
This still has issues on i686 Windows that I need to fix...

adamreichold:
> Would be nice to remove this then, but probably not in this PR I guess as it is rather unrelated to the change at hand.

Created #98 for that.

AndrewGaspar:
This will make #44 defunct.

build.rs
```diff
@@ -1,6 +1,26 @@
+use std::env::var;
+
 fn main() {
-    if cfg!(windows) {
-        // Adds a cfg to identify MS-MPI. This should perhaps be a more robust check in the future.
+    let is_msmpi = if let Ok(library) = build_probe_mpi::probe() {
```
adamreichold, line 4: This might be a bit on the verbose side, e.g.
```
build_probe_mpi::probe().map_or(false, |library| library.version == "MS-MPI")
```

src/collective.rs
```diff
@@ -1786,15 +1786,15 @@ unsafe impl AsRaw for UnsafeUserOperation {
...
+#[cfg(not(all(msmpi, target_arch = "x86")))]
 pub type UnsafeUserFunction =
     unsafe extern "C" fn(*mut c_void, *mut c_void, *mut c_int, *mut ffi::MPI_Datatype);
 
 /// A raw pointer to a function that can be used to define an `UnsafeUserOperation`.
 ///
-/// MS-MPI uses "stdcall" rather than "C" calling convention. These are actually equivalent on
-/// 64-bit Windows, but diverge on 32-bit Windows.
-#[cfg(msmpi)]
+/// MS-MPI uses "stdcall" rather than "C" calling convention. "stdcall" is ignored on x86_64
```
adamreichold, line 1795: I am slightly puzzled: The build script disables user operations on 32-bit Windows but this does provide a special case of the `UnsafeUserFunction` type. Is this type alias used for something else besides user operations? Could we remove it entirely instead?

adamreichold:
This might be an unpopular opinion, but `libffi` seems to be used here to add functionality that MPI does not have itself: Passing context into the user operations, i.e. making a function pointer out of a closure instance via runtime code gen.

Wouldn't it be much simpler to mirror MPI in the Rust API, i.e. take only function pointers like `fn(DynBuffer, DynBufferMut)` (which can also be created by non-capturing closures) and handle changing the calling conventions and wrapping the arguments.

If users want access to state in their operations, they'll have to provide access via global data manually. While this is slightly more work, it should avoid any doubts on the safety of the construction and make the construction of a self-referential type unnecessary.

src/collective.rs
AndrewGaspar: `UnsafeUserFunction` doesn't require `user-operations`. `user-operations` provides "safe" MPI OPs. So this still lets 32-bit Windows use the unsafe interface.

AndrewGaspar:
> Wouldn't it be much simpler to mirror MPI in the Rust API, i.e. take only function pointers like fn(DynBuffer, DynBufferMut) (which can also be created by non-capturing closures) and handle changing the calling conventions and wrapping the arguments.

So as far as I know, there's no way to pass a function as a generic type argument like you can in C++. So there's no way to statically get the function address in a way that you could call it from a context-less lambda. Instead, you would need to wrap the function in a static trait impl, and then pass the type of that trait instead. I think you can easily do this with sufficient macros: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c19d7c9d31677ffe16f7b6b1eddd616e

You still need to pass the "callback" around as a regular argument, because that's the only way to get the type. It's otherwise anonymous as it's declared in a private scope by a macro.

I think this is as good idea - I'll provide a safe raw user operations feature that does this and supplant the unsafe version.

> If users want access to state in their operations, they'll have to provide access via global data manually. While this is slightly more work, it should avoid any doubts on the safety of the construction and make the construction of a self-referential type unnecessary.

Well, that's why the `user-operations` feature exists. It not only avoids "slightly more work", it actually makes certain things possible that otherwise wouldn't be - imagine instantiating some variable number of operations that only differ by some constant. There's not enough context in the MPI_Op callback to determine which instantiation is associated with which constant. Using `libffi` resolves that problem.

adamreichold:
> I think this is as good idea - I'll provide a safe raw user operations feature that does this and supplant the unsafe version.

Indeed, the playground you linked looks like a workable way to translate the calling conventions when the signature is known up front. And it does without involving the question whether to add context or not.

> There's not enough context in the MPI_Op callback to determine which instantiation is associated with which constant. Using libffi resolves that problem.

What I mean is that it does two things at once here: Changing the calling convention and providing runtime code gen. As you have shown above, we strictly need libffi only for the second part. So I mainly wonder whether there are simpler and orthogonal ways of getting the code gen functionality.

src/collective.rs
adamreichold: Understood.

AndrewGaspar:
> So I mainly wonder whether there are simpler and orthogonal ways of getting the code gen functionality.

I'd love to be proven wrong - libffi by its nature is not automatically portable. However, I think the MPI_Op interface is just inherently limited in this regard.

AndrewGaspar:
> Changing the calling convention and providing runtime code gen.

Well, I'm not sure that's what people typically use it for. It's mainly for the closure capabilities.

adamreichold:
> Well, I'm not sure that's what people typically use it for. It's mainly for the closure capabilities.

Exactly, but our issues are with the calling convention part, aren't they?

I have though about this some more and I guess doing what libffi does will always end up in code gen and hence will platform-specific. So I guess the whole point is moot and we should just stick to libffi and try to the fix the calling convention issues there. 

In that case however,

> I'll provide a safe raw user operations feature that does this and supplant the unsafe version.

does not seem to helpful IMHO as it will only complicate things and provide multiple redundant paths to achieve the same thing and I would therefore recommend not doing it.

Some questions on technicalities from looking over the code:
- Why does `UserOperation` try so hard to erase the closure type introducing the `Erased` trait instead of storing something like `Box<ClosureAnchor<F>>`?
- I think `ClosureAnchor` could use `MaybeUninit` instead of `Option` around the `ffi_closure` to avoid the extra tag? (It involves unsafe code either way.)

adamreichold:
> So I guess the whole point is moot and we should just stick to libffi and try to the fix the calling convention issues there.

@AndrewGaspar I am very sorry for side tracking the whole with what is in retrospect mainly me not sufficiently understanding the problem domain.

AndrewGaspar:
> Why does UserOperation try so hard to erase the closure type introducing the Erased trait instead of storing something like Box<ClosureAnchor<F>>?

Here's the original discussion - https://github.com/rsmpi/rsmpi/pull/25#discussion_r118565057

I think I generally agree that preserving the type isn't all that useful and complicates usage without removing any overhead.

AndrewGaspar:
> I think ClosureAnchor could use MaybeUninit instead of Option around the ffi_closure to avoid the extra tag? (It involves unsafe code either way.)

My testing shows that there is no size difference between `Closure4` and `Option<Closure4>`, and I don't think the branch on drop is all that impactful.

AndrewGaspar:
@adamreichold look good?

adamreichold:
> @adamreichold look good?

Yes, this looks good. #97 is unfortunate but I guess we will need to do nothing but change the feature flags when libffi handles this? (I guess 32-bit Windows is also not the most widely used platform for distributed simulations.)

AndrewGaspar:
We might want to wait for libffi to update before releasing 0.6, though... they also broke other architectures that ARE used, like aarch64 and powerpc64le.

https://github.com/tov/libffi-rs/pull/17

AndrewGaspar:
I ended up adding support for 32-bit Windows with MS-MPI. Yeah, it's a pretty questionable use-case, but I figured I might as well support it if it's going to be the default. Finally fixed a pet-peeve for me where the Windows headers weren't found by bindgen unless you built from a Visual Studio developer command prompt, and even then it was broken for cross-compiling. This change fixes that.

Fixes #57.
Fixes #97.

mpi-sys/build.rs
```diff
@@ -55,6 +55,22 @@ fn main() {
         builder = builder.clang_arg(format!("-I{}", dir.display()));
     }
 
+    // Get the same system includes as used to build the "rsmpi" lib. This block only really does
+    // anything when targeting msvc.
+    if let Ok(compiler) = compiler {
+        let include_env = compiler
+            .env()
+            .iter()
+            .filter(|(key, _)| key == "INCLUDE")
```
adamreichold, line 64: Isn't this `.find(...)`?

src/collective.rs
```diff
@@ -1687,22 +1687,55 @@ impl<'a> UserOperation<'a> {
...
+            Type::pointer(), // int32_t *
+            Type::pointer(), // MPI_Datatype *
+        ];
+        #[allow(unused_mut)]
+        let mut cif = Cif::new(args, Type::void());
+        // MS-MPI uses "stdcall" calling convention on 32-bit x86
+        #[cfg(all(msmpi, target_arch = "x86"))]
+        cif.set_abi(libffi::raw::ffi_abi_FFI_STDCALL);
+
+        unsafe extern "C" fn trampoline<
```
adamreichold, line 1713: If we do have the trampoline under control, could we fold the landing pad into it to simplify the control flow a bit?

src/collective.rs
```diff
@@ -1687,22 +1687,55 @@ impl<'a> UserOperation<'a> {
...
+
         // must box it to prevent moves
         let mut anchor = Box::new(ClosureAnchor {
             rust_closure: move |invec, inoutvec, len, datatype| unsafe {
                 user_operation_landing_pad(&function, invec, inoutvec, len, datatype)
             },
             ffi_closure: None,
         });
+
+        let args = vec![
```
adamreichold, line 1701: `Cif::new` seems to want an `ExactSizeIterator` eventually, so maybe an array would here to avoid the temporary allocation?

mpi-sys/build.rs
```diff
@@ -21,26 +21,26 @@ fn main() {
...
             builder.include(inc);
         }
 
-        builder.compile("librsmpi.a");
-
         // Adds a cfg to identify MS-MPI
         println!("cargo:rustc-cfg=msmpi");
     } else {
+        // Use `mpicc` wrapper on Unix rather than the system C compiler.
         env::set_var("CC", "mpicc");
```
adamreichold, line 38: As the MPICH and OpenMPI builds on Travis failed, I think this needs to be set before calling `cc::Build::new`?

src/collective.rs
AndrewGaspar: Unfortunately, fixed size arrays don't implement `IntoIterator`. `IntoIterator` is only implemented for references to a fized size array.

I used `smallvec` instead, which we already have a dependency on.

adamreichold: > Unfortunately, fixed size arrays don't implement IntoIterator.

`I: Iterator` implies `I: IntoIterator`, so you can explicitly create the iterator using e.g. `[1,2,3].iter().copied()`, c.f. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e4ed29a91395aca8ef007a100d6f7ecc

AndrewGaspar: `Type` isn't `Copy`.

mpi-sys/build.rs
AndrewGaspar: I make this mistake all the time.

src/collective.rs
AndrewGaspar: I guess `Clone` is zero-allocation for `Type::pointer`, so I could do `cloned()` instead.

src/collective.rs
adamreichold: I was actually thinking of putting the argument unpacking logic directly into `trampoline` (where other argument unpacking is already done) and store the user-supplied function directly in `rust_closure`.

adamreichold:
> We might want to wait for libffi to update before releasing 0.6, though... they also broke other architectures that ARE used, like aarch64 and powerpc64le.
> 
> [tov/libffi-rs#17](https://github.com/tov/libffi-rs/pull/17)

This looks good to me now. Maybe should make a note on the 0.6 release issue due to the above?

AndrewGaspar:
> Maybe should make a note on the 0.6 release issue due to the above?

Hm... I worry about confusing people. Most people updating to "0.6" would not previously have a dependency on libffi "1.0.0", and so they would almost certainly get libffi-sys "1.1.0" by default when the libffi dependency updates. I think it creates more trouble than it's worth.

I'm working adding on a CHANGELOG.

I'm going to restart the tests. Pretty sure it's spurious. Unfortunately, I don't have permission to restart jobs on the Travis CI. I'd like to switch to GitHub Actions at some point.

PR #101, opened 2020-11-18
Cleanup clippy prior to release

AndrewGaspar:
This is mostly clippy-related cleanup, but I did make a few API changes that I encountered along the way:

I removed `unsafe` from the `UnsafeUserOperation` constructors. I didn't see any reason why constructing an `UnsafeUserOperation` would be unsafe. The only unsafety should be in the actual reduction function itself, which must be marked `unsafe`.

I removed the `unsafe` designator from the `pointer` and `pointer_mut` functions in the `Pointer` and `PointerMut` traits, respectively. Those traits are already marked as `unsafe`, and there's no reason that merely getting the pointer to an object should be unsafe.

I removed some helper functions from the `Partitioned` trait. Nobody should ever be implementing the `Partitioned` trait anyway.


src/collective.rs
```diff
@@ -1846,9 +1849,11 @@ impl UnsafeUserOperation {
     /// # Standard section(s)
     ///
     /// 5.9.5
-    pub unsafe fn new(commute: bool, function: UnsafeUserFunction) -> Self {
```
bsteinb, line 1849: This one is tricky. I agree that the construction of UnsafeUserOperation per se does not perform `unsafe` operations. The problem lies in the `*reduce*` methods in `collective.rs` which in combination with `UnsafeUserOperation` allow you to call `unsafe` functions without introducing `unsafe` blocks in user code. (The definition of `function` could be part of a library.)

Strictly speaking, the `*reduce*` methods should be marked `unsafe` then, but only if they are passed `unsafe` operations. Alas, there is no such thing as parametric `unsafe`, I think.

adamreichold: I also think that the constructor should be kept unsafe because ignore the whole MPI part around it, this is a way to call an unsafe function (`UnsafeUserFunction`) without an `unsafe` block. Hence the constructor itself must require `unsafe` to assert that calling the reduction functions will be safe. (The `unsafe` where the function is implemented is the kind that tells me that there are additional invariants that need manual checking so there must be another `unsafe` which asserts that these invariants are fulfilled.)

adamreichold (changes requested):
I think the change to `UnsafeUserOperation::new` needs to be discussed further. I also wonder whether we should add a Clippy job to the CI?

src/collective.rs
AndrewGaspar: OK, I can accept that explanation. I'll remove that change.

AndrewGaspar:
> I think the change to `UnsafeUserOperation::new` needs to be discussed further. I also wonder whether we should add a Clippy job to the CI?

I made a note over on #91.

adamreichold (approved):
Looks good to me now.

PR #106, opened 2021-03-11
Fix derive Equivalence for arrays + formatting

iwahbe:
Allows the derive macro to be used on static arrays with non `{integer}`
type constants, Specifically, `usize`. I include a test to prevent
regression.

I made 2 cosmetic changes in collective.rs and examples/struct.rs to
allow `rsmpi` to compile without warnings.

### DIFF
--- a/mpi-derive/src/lib.rs
+++ b/mpi-derive/src/lib.rs
@@ -43,7 +43,12 @@ fn equivalence_for_tuple_field(type_tuple: &syn::TypeTuple) -> TokenStream2 {
 fn equivalence_for_array_field(type_array: &syn::TypeArray) -> TokenStream2 {
     let ty = equivalence_for_type(&type_array.elem);
     let len = &type_array.len;
-    quote! { &::mpi::datatype::UncommittedUserDatatype::contiguous(#len as i32, &#ty) }
+    // We use the len block to ensure that len is of type `usize` and not type
+    // {integer}. We know that `#len` should be of type `usize` because it is an
+    // array size.
+    quote! { &::mpi::datatype::UncommittedUserDatatype::contiguous(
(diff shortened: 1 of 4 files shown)


mpi-derive/src/lib.rs
```diff
@@ -43,7 +43,7 @@ fn equivalence_for_tuple_field(type_tuple: &syn::TypeTuple) -> TokenStream2 {
 fn equivalence_for_array_field(type_array: &syn::TypeArray) -> TokenStream2 {
     let ty = equivalence_for_type(&type_array.elem);
     let len = &type_array.len;
-    quote! { &::mpi::datatype::UncommittedUserDatatype::contiguous(#len, &#ty) }
+    quote! { &::mpi::datatype::UncommittedUserDatatype::contiguous(#len as i32, &#ty) }
```
AndrewGaspar, line 46: Could you do something like this instead:

```suggestion
    quote! { &::mpi::datatype::UncommittedUserDatatype::contiguous(#len.try_into().expect(&format!("rsmpi derive: Array size too large for MPI_Datatype: {}", #len)), &#ty) }
```

It's unlikely anybody would ever have an array size that big, but might as well fail loudly if they do.

AndrewGaspar (changes requested):
Thank you for this - very reasonable fix! I have one small change request.

mpi-derive/src/lib.rs
iwahbe: I did something similar. I used a static `&str` instead of a format invocation because I didn't want to make the allocation from `format` each time `equivalent_datatype` is called. Otherwise it has the same effect.

AndrewGaspar: I would guess that the actual allocation would be elided except in failure cases since it can be statically determined if the constant value is convertible to an `i32`. But that's fine - looks good!

AndrewGaspar (approved):

PR #110, opened 2021-11-09
Implement Buffer and BufferMut for Vec

hhirtz:
To allow `&vec` instead of `vec.as_slice()`.


jedbrown:
Thanks for this PR. Can you rebase on `main` (so CI can run) and add a test (perhaps by modifying/extending a current example)?

hhirtz:
Rebased and modified two examples for the `Vec` impl and another for the array impl

PR #113, opened 2022-02-08
Make `Communicator` and `Group` traits "object safe"

### LINKED ISSUE #108: Some traits are not object safe, but could be.
Hi,

I just started using this library, so I'm not very familiar yet, but I encountered a problem. I realized that the `Communicator` trait is not currently "object safe," so something like the following will fail to compile.

```rust
fn foo(comm: &dyn Communicator) { }
```

This is something I need for my application as I don't want to care what specific type of communicator I have, and I only need a small subset of the methods like `size()`, `rank()`, and `as_raw()`.

It seems to be a fairly easy fix as all that is needed is to add the trait bound `Self: Sized` to the methods that use generics. 

I haven't looked around much, but I also found that the same thing applies to the `Group` trait.

Is this a change you would want?

ZackJorquera:
This PR makes the `Communicator` and `Group` traits "object safe," it will allow you to write things like:

```rust
fn foo(comm: &dyn Communicator) { }
fn bar(gp: &dyn Group) { }
```

fixes #108

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -337,6 +337,7 @@ pub trait Communicator: AsRaw<Raw = MPI_Comm> {
     fn compare<C: ?Sized>(&self, other: &C) -> CommunicatorRelation
     where
         C: Communicator,
+        Self: Sized,
     {
         unsafe {
             with_uninitialized(|cmp| ffi::MPI_Comm_compare(self.as_raw(), other.as_raw(), cmp))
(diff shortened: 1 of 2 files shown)


jedbrown:
Thanks; pipeline is running. Would you be able to add a simple test in this PR?

ZackJorquera:
I could probably do that. How and where should I add it?

jedbrown:
Add or modify an existing example (those are run with 2..4 processes) or add in `tests/` (which will be run sequentially, but that's faster and probably sufficient). I'd just like a test so we don't accidentally lose object safety in the future.

jedbrown (approved):
Great, thanks.

Issue #114, opened 2022-02-11
macOS compiling error


jianshu93:
Hi team,

I have the following error when compiling on MacOS:

error[E0432]: unresolved import `crate::raw::ffi_type_longdouble`
   --> /Users/jianshuzhao/.cargo/registry/src/github.com-1ecc6299db9ec823/libffi-1.0.1/src/low.rs:140:13
    |
140 |     pub use crate::raw::ffi_type_longdouble as longdouble;
    |             ^^^^^^^^^^^^-------------------^^^^^^^^^^^^^^
    |             |           |
    |             |           help: a similar name exists in the module: `FFI_TYPE_LONGDOUBLE`
    |             no `ffi_type_longdouble` in `raw`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `libffi` due to previous error

thanks,

Jianshu

bsteinb:
I can reproduce the issue on macOS on Apple Silicon, macOS on Intel is not affected. However, the issue is in the `libffi` crate. It has been fixed recently, see https://github.com/tov/libffi-rs/pull/51, but there has been no release of `libffi` on crates.io that includes the fix.

jianshu93:
Thanks! You can edit the Cargo.toml in rsmpi to use local path, where you download the newest libffi-rs: libffi = { path = "../libffi-rs", optional = true }


Jianshu

jianshu93:
Or you can always use: libffi = { git = "https://github.com/tov/libffi-rs", optional = true}
I have no errors on apple M1. And finally I can test rust MPI on my new device!!!

Jianshu

bsteinb:
Fixed in #116

PR #116, opened 2022-02-14
libffi: bump version to fix aarch64 issue #114

jedbrown:
@jianshu93 Can you confirm that this works for you?


jianshu93:
yes confirmed. 

Jianshu

PR #122, opened 2022-07-28
Multiple Request Completion 2

jtronge:
Hello, I've been working with @hppritcha to attempt adding multiple request completion support. This is partly based on @AndrewGaspar's original PR #36 but is slightly different. Based on some of the discussion in #29, I experimented with attaching references to each request as a way to allow the user to immediately perform some computation on the data as soon as the request has been waited on.

Using the modified request object as a basis I added a RequestCollection object, similar to the former PR, but including support for attached references. Requests can be added to the collection and then waited on using the wait_*() or test_*() methods. To ensure that requests have completed, I included a new `multiple_scope()` function which passes in both a reference to a scope object and a request collection.

### DIFF
--- a/examples/immediate_multiple_requests.rs
+++ b/examples/immediate_multiple_requests.rs
@@ -1,0 +1,152 @@
+//! Example using request handlers.
+use mpi;
+use mpi::point_to_point::Status;
+use mpi::traits::*;
+
+const COUNT: usize = 256;
+
+fn main() {
+    let universe = mpi::initialize().unwrap();
+    let world = universe.world();
+    let size = world.size();
+    let rank = world.rank();
+
+    let x: i32 = 100;
+    let mut y: i32 = 0;
+
+    mpi::request::scope(|scope| {
+        let sreq = world.this_process().immediate_send(scope, &x);
+        let rreq = world.any_process().immediate_receive_into(scope, &mut y);
+        let result = rreq.wait_for_data();
+        assert_eq!(*result, x);
+        sreq.wait();
+    });
+
+    // Test wait_any()
+    let mut result: Vec<i32> = vec![0; COUNT];
+    mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
+        let prev_proc = if rank > 0 { rank - 1 } else { size - 1 };
+        let next_proc = (rank + 1) % size;
+        for _ in 0..result.len() {
+            let sreq = world
+                .process_at_rank(next_proc)
+                .immediate_send(scope, &rank);
+            coll.add(sreq);
+        }
+        for val in result.iter_mut() {
+            let rreq = world
+                .process_at_rank(prev_proc)
+                .immediate_receive_into(scope, val);
+            coll.add(rreq);
+        }
+        let mut send_count = 0;
+        let mut recv_count = 0;
+        while coll.incomplete() > 0 {
+            let (_, _, result) = coll.wait_any().unwrap();
+            if *result == rank {
+                send_count += 1;
+            } else {
+                recv_count += 1;
+            }
+        }
+        assert_eq!(send_count, COUNT);
+        assert_eq!(recv_count, COUNT);
+    });
+
+    let mut result: Vec<i32> = vec![0; COUNT];
+    // Test wait_some()
+    mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
+        let prev_proc = if rank > 0 { rank - 1 } else { size - 1 };
+        let next_proc = (rank + 1) % size;
+        for _ in 0..result.len() {
+            let sreq = world
+                .process_at_rank(next_proc)
+                .immediate_send(scope, &rank);
+            coll.add(sreq);
+        }
+        for val in result.iter_mut() {
+            let rreq = world
+                .process_at_rank(prev_proc)
+                .immediate_receive_into(scope, val);
+            coll.add(rreq);
+        }
+        let mut send_count = 0;
+        let mut recv_count = 0;
+        let mut some_buf: Vec<Option<(usize, Status, &i32)>> = vec![None; 2 * COUNT];
+        while coll.incomplete() > 0 {
+            let count = coll.wait_some(&mut some_buf);
+            println!("wait_some(): {} request(s) completed", count);
+            assert!(some_buf.iter().any(|elm| elm.is_some()));
+            for elm in some_buf.iter() {
+                if let Some((_, _, result)) = elm {
+                    if **result == rank {
+                        send_count += 1;
+                    } else {
+                        recv_count += 1;
+                    }
+                }
+            }
+        }
+        assert_eq!(send_count, COUNT);
+        assert_eq!(recv_count, COUNT);
+    });
+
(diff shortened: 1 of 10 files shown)


hppritcha:
could someone kick off the CI for this PR @bsteinb @jedbrown  thanks.

examples/immediate_multiple_requests.rs
```diff
@@ -0,0 +1,152 @@
...
+            assert!(some_buf.iter().any(|elm| elm.is_some()));
+            for elm in some_buf.iter() {
+                if let Some((_, _, result)) = elm {
+                    if **result == rank {
+                        send_count += 1;
+                    } else {
+                        recv_count += 1;
+                    }
+                }
+            }
```
jedbrown, line 88: What happens if 2 requests complete the first time around this loop and then only 1 completes? I don't think `some_buf` is reset to `None` so it'll process `some_buf[1]` again the second time around the loop.

jedbrown (commented):
Thanks for working on this. I see some `cargo fmt` noise and have some questions about how to expose the completions.

examples/immediate_multiple_requests.rs
jtronge: Yeah you're right. I should either reset the vector here before calling `wait_some` or have `wait_some` ensure that unused slots are `None`.

src/request.rs
```diff
@@ -463,3 +523,263 @@ where
...
+                assert!(is_null(self.requests[idx]));
+                if let Some(data) = self.data[idx].take() {
+                    let status = unsafe { self.statuses[idx].assume_init() };
+                    let status = Status::from_raw(status);
+                    let _ = elm.insert((idx, status, data));
+                }
+            }
+        }
+        count
+    }
```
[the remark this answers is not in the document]
jtronge: You're probably right, that would be easier to work with. I guess I was thinking about using a slice here instead of limiting it to a `Vec`. I assume you mean removing the `Option` and passing a `&mut Vec<(usize, Status, &'a D)>`?

jtronge:
Thanks for taking a look at this. I'll fix the lints and take a closer look at the `wait_some`/`test_some` code at the start of next week.

src/request.rs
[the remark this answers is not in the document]
jedbrown: Yeah, that seems much easier to use correctly and it still doesn't imply any allocation in a hot loop. (MPI itself may allocate in these patterns, but we won't be adding any mandatory allocation like there would be if we were to return a new `Vec` instead of using the provided `&mut Vec`.

examples/immediate_multiple_requests.rs
jedbrown: This would get simpler and this mistake will go away if you move to the `&mut Vec` model.

jedbrown:
I'd like to release version 0.6 soon so other crates can publish. Do you think this should be part of 0.6 and if so, how long do you anticipate needing?

jedbrown:
Heads up that there are some new clippy warnings. This is clear on main.
```console
$ cargo clippy -- --allow clippy::missing_safety_doc
```

jtronge:
Ok, thanks. I think I managed to fix all of the warnings. I've updated the code and examples to use `&mut Vec`.

examples/immediate_multiple_requests.rs
```diff
@@ -0,0 +1,151 @@
...
+        let mut recv_count = 0;
+        let mut some_buf = vec![];
+        let mut finished = 0;
+        while coll.incomplete() > 0 {
+            coll.wait_some(&mut some_buf);
+            println!("wait_some(): {} request(s) completed", some_buf.len());
+            finished += some_buf.len();
+            assert_eq!(coll.incomplete(), 2 * COUNT - finished);
+            for (_, _, result) in some_buf.iter() {
+                if **result == rank {
```
jedbrown, line 83: What would you think of this?
```suggestion
            for &(_, _, result) in some_buf.iter() {
                if *result == rank {
```

examples/immediate_multiple_requests.rs
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+        assert_eq!(send_count, COUNT);
+        assert_eq!(recv_count, COUNT);
+    });
+
+    let mut result: Vec<i32> = vec![0; COUNT];
+    // Test wait_all()
+    mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
+        let prev_proc = if rank > 0 { rank - 1 } else { size - 1 };
```
jedbrown, line 97: This is the standard pattern in MPI codes, but I see rsmpi uses the one you have in other places so I can update them all later.
```suggestion
        let prev_proc = (rank - 1 + size) % size;
```

src/lib.rs
```diff
@@ -179,6 +179,8 @@ pub type Count = c_int;
 pub type Tag = c_int;
 /// An address in memory
 pub type Address = MPI_Aint;
+/// Rank of each process.
+pub type Rank = c_int;
```
jedbrown, line 183: Is it intentional that this duplicates the definition in `src/topology/mod.rs`?

jedbrown (commented):
Looks great, just a couple comments to consider. And I'll run a pipeline.

examples/immediate_multiple_requests.rs
jtronge: That would be more readable. That double dereference is a little confusing to look at.

jedbrown:
I don't think this is from your branch (it [passed on re-running](https://github.com/rsmpi/rsmpi/runs/7658914760?check_suite_focus=true)), but do you understand the intermittent failure here? It looks like perhaps `std::panic::catch_unwind` not working correctly.

https://github.com/rsmpi/rsmpi/runs/7658512360?check_suite_focus=true#step:8:120

jedbrown:
Many thanks for this important contribution. I'll merge now.

jtronge:
> I don't think this is from your branch (it [passed on re-running](https://github.com/rsmpi/rsmpi/runs/7658914760?check_suite_focus=true)), but do you understand the intermittent failure here? It looks like perhaps `std::panic::catch_unwind` not working correctly.
> 
> https://github.com/rsmpi/rsmpi/runs/7658512360?check_suite_focus=true#step:8:120

Hmmn, I'm not sure. I've run into a couple strange errors like this when Rust panics, or the c code that rust is calling aborts. It looks like openmpi is failing here and I'm not sure where the Rust code is handling this, or if it even is.

jedbrown:
There is output from where Rust panics (that's supposed to be caught), but the `unlink` error is quite strange. It'll be something to watch, especially if we find a way to reproduce it.

jtronge:
Talking with @hppritcha, he pointed me to this issue: https://github.com/open-mpi/ompi/issues/7393. Seems like we might want to set `TMPDIR` for macOS at least. He also suggested adding a sleep between tests.

jedbrown:
Thanks, I haven't seen this failure since https://github.com/rsmpi/rsmpi/pull/125.

I've been seeing non-deterministic MSMPI failures recently, but I don't think they're related to recent changes to rsmpi.

PR #124, opened 2022-08-01
Improve pkg-config detection and support Cray

jedbrown:
This is tested with `PrgEnv-aocc` on [NERSC Perlmutter](https://www.nersc.gov/systems/perlmutter/). It detects the environment and finds it by default based on `CRAY_MPICH_DIR`. It did not work with `PngEnv-cray`, but I think thsi needs to be fixed in [clang-sys](https://github.com/KyleMayes/clang-sys/issues/143). I expect it will work with upstream clang on Cray.

I also fixed some incomplete propagation of `mpicc` and include paths that previously caused `rsmpi.c` to be compiled using default paths, and thus fail or build for the wrong MPI when intending to use a non-default MPI.

@skailasa or @tbetcke Would you be able to test this on one of your systems?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,9 +6,9 @@ authors = [
     "Andrew Gaspar <andrew.gaspar@outlook.com>"
 ]
 description = "Message Passing Interface bindings for Rust"
-homepage = "https://github.com/bsteinb/rsmpi"
-repository = "https://github.com/bsteinb/rsmpi"
-documentation = "http://bsteinb.github.io/rsmpi/"
+homepage = "https://github.com/rsmpi/rsmpi"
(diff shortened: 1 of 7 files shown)


jedbrown:
Also tested on OLCF Crusher with `PrgEnv-amd` (but not `PrgEnv-cray`).
```console
$ cc --version
AMD clang version 14.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.1.0 22114 5cba46feb6af367b1cafaa183ec42dbfb8207b14)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /opt/rocm-5.1.0/llvm/bin
```
I'll go ahead and merge, and we can refine later if needed.

PR #126, opened 2022-08-04
versions and pkg-config

jedbrown:
Finer grained pkg-config selection, documentation of version selection, update versions, and housekeeping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,8 @@ fn main() {
     let size = world.size();
     let rank = world.rank();
 
-    let next_rank = if rank + 1 < size { rank + 1 } else { 0 };
-    let previous_rank = if rank > 0 { rank - 1 } else { size - 1 };
+    let next_rank = (rank + 1) % size;
+    let previous_rank = (rank - 1 + size) % size;
 
     let msg = vec![rank, 2 * rank, 4 * rank];
     mpi::request::scope(|scope| {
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -78,8 +87,8 @@ fn main() {
     let size = world.size();
     let rank = world.rank();
 
-    let next_rank = if rank + 1 < size { rank + 1 } else { 0 };
-    let previous_rank = if rank > 0 { rank - 1 } else { size - 1 };
+    let next_rank = (rank + 1) % size;
+    let previous_rank = (rank - 1 + size) % size;
```
adamreichold, line 91: This might be a common idiom, but it is also a rather opaque transformation which I could image a compiler could do on its own. If it is important to use it, I would suggest putting into simple `#[inline] fn` and making them part of this crate's API.

jedbrown: Hmm, there's no performance concern at all and this is specific to a 1D periodic point distribution. The Cartesian topology interfaces are supported (see `examples/cartesian.rs`) and can be used for 1D distributions, but I think `(rank + 1) % size` is more transparent to readers that there is no magic.

jedbrown: I'm going to resolve and merge this for now. I appreciate your input and would be happy to discuss further or merge an MR, but I struggle to see the value in the context of examples meant to be illustrative of what a user can build.

PR #128, opened 2022-08-13
Allow receiving into `MaybeUninit<T>` where `T: Equivalence`

juntyr:
This PR implements `Equivalence` for `MaybeUninit<T>` where `T: Equivalence` by forwarding the impl to `T`. This is useful, for instance, when receiving into the spare capacity of a `Vec<T>`, which is returned as `&mut [MaybeUninit<T>]`, in order to avoid having uninitialised items.

### DIFF
--- a/src/datatype.rs
+++ b/src/datatype.rs
@@ -213,14 +213,6 @@ equivalent_system_datatype!(usize, ffi::RSMPI_UINT64_T);
 #[cfg(target_pointer_width = "64")]
 equivalent_system_datatype!(isize, ffi::RSMPI_INT64_T);
 
-unsafe impl<T: Equivalence> Equivalence for mem::MaybeUninit<T> {
-    type Out = T::SystemDatatype;
-
-    fn equivalent_datatype() -> Self::Out {
-        T::equivalent_datatype()
-    }
-}
-
 /// A user defined MPI datatype
 ///
 /// # Standard section(s)
@@ -796,6 +788,16 @@ where
     }
 }
 
+unsafe impl<T> AsDatatype for [mem::MaybeUninit<T>]
+where
+    T: Equivalence,
+{
+    type Out = <T as Equivalence>::Out;
+    fn as_datatype(&self) -> Self::Out {
+        <T as Equivalence>::equivalent_datatype()
+    }
+}
+
 unsafe impl<T> AsDatatype for Vec<T>
 where
     T: Equivalence,
@@ -897,6 +899,17 @@ where
     }
 }
 
+unsafe impl<T> Collection for [mem::MaybeUninit<T>]
+where
+    T: Equivalence,
+{
+    fn count(&self) -> Count {
+        self.len()
+            .value_as()
+            .expect("Length of slice cannot be expressed as an MPI Count.")
+    }
+}
+
 unsafe impl<T> Collection for Vec<T>
 where
     T: Equivalence,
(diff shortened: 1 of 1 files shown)


jedbrown (commented):
Thanks. Is this going to allow sending from a `MaybeUninit<T>`, not just receiving into?

Could you add a test or example?

MomoLangenstein:
> Is this going to allow sending from a `MaybeUninit<T>`, not just receiving into?

Hmmm, good point. I only want to allow receiving into it, not sending from it. Would it be better to add extra impls for `BufferMut` to allow sending from `MaybeUninit<T>` as well? What about cases where the implementer also impls `Buffer`?

> Could you add a test or example?

Yes, once we've settled on a design I'll add some :)

jedbrown:
I'd think we want to ensure that the user *cannot* send from `MaybeUninit<T>`. Once received, the written buffer would be initialized. I could imagine an interface in which receives return a suitably typed slice from the provided buffer, but that might have undesirable side-effects.

What is your main objective in this PR? Like could you write a brief example in which the current interface is worse? See also #79, which is related.

MomoLangenstein:
In my use case, I previously wrote something akin to:

```rust
let receive_start = receiving_buffer.len();

#[allow(clippy::uninit_vec)]
// Safety: The uninitialised `number_items` items are initialised in the
//         following `matched_receive_into` call
unsafe {
    receiving_buffer.reserve(number_items);
    receiving_buffer.set_len(receive_start + number_items);

    msg.matched_receive_into(&mut receiving_buffer[receive_start..]);
}
```

where `receiving_buffer` is a `Vec` and `number_items` refers to the number of items we will receive from `msg`. Unfortunately, this code claims that the last `number_items` items are initialised (by calling `set_len`) before we have written the new items into the vec. Hence, the code has undefined behaviour.

With the new impls, I can instead write:

```rust
let receive_start = receiving_buffer.len();
receiving_buffer.reserve(number_items);

msg.matched_receive_into(receiving_buffer.spare_capacity_mut()[..number_items]);

// Safety: The uninitialised `number_items` items were just initialised
//         in the `matched_receive_into` call
unsafe {
    receiving_buffer.set_len(receive_start + number_items);
}
```

which only marks the newly received items as initialised once we have actually received them into the (uninitialised) spare capacity of our buffer `Vec`.

jedbrown:
Presser is a new package that might help us deal with `MaybeUninit` correctly, similarly to its [`copy_into_maybe_uninit_slice`](https://docs.rs/presser/latest/presser/fn.copy_into_maybe_uninit_slice.html).
```rs
pub fn copy_into_maybe_uninit_slice<'a, T>(src: &[T], dst: &'a mut [MaybeUninit<T>]) -> &'a mut [T]
where
    T: Copy,
{
    let uninit_src: &[MaybeUninit<T>] =
        // SAFETY: &[T] and &[MaybeUninit<T>] have the same layout
        unsafe { &*(src as *const [T] as *const [MaybeUninit<T>]) };

    dst.copy_from_slice(uninit_src);

    // SAFETY: Valid elements have just been copied into `this` so it is initialized
    unsafe { &mut *(dst as *mut [MaybeUninit<T>] as *mut [T]) }
}
```

@MomoLangenstein I think to merge your PR, we'll need it to have tests and to discuss safety. It seems undesirable to me for the caller to have to do the [`slice_assume_init_mut()`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_mut) (or the messy unsafe cast above because this is nightly-only).

Issue #118, opened 2022-03-30
Example project of integrating rsmpi into mpi4py with maturin


tbetcke:
Hi,

I have created a small demonstration project of how to pass a communicator from mpi4py into rsmpi using automatically generated cffi bindings from `maturin`. 

The code is available at:

https://github.com/betckegroup/mpi4py_with_rsmpi

The difficulty was passing the communicator. If we directly want to pass a pointer to MPI_Comm we would have to include the "mpi.h" header file into the maturin build process, which caused us quite a few errors. So we ended up converting a pointer to MPI_Comm into a uintptr_t and passed this to a Rust usize, from where we could convert this back to a pointer to MPI_Comm. Since maturin only sees a C interface that takes an integer this is straightforward to generate bindings for and is handled automatically by maturin.

If you think it useful I would be happy for a link to this demonstration project to be added to the rsmpi documentation.

jedbrown:
Thanks for building this example. I think we could add a section to the README (or other place in the docs) with recommended demonstrations. That could be further split into simple demos and featured production apps/libraries. In this case, I'd almost say we can include it in the rsmpi repo directly, but that might require some labor to ensure it's tested in CI. Feel free to make a PR either way (just docs or making it a demo that's tested on at least one platform).

jedbrown:
Resolved by #129.

PR #129, opened 2022-08-13
Add documentation link to example project for using rsmpi with mpi4py

tbetcke:
As suggested in #118 I have created a documentation link to an example project that uses mpi4py with rsmpi. If there are more example bindings to other langauges this may move into a subdirectory of rsmpi. But given there is only one example it might be easiest to just add a documentation link for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ See files in [examples/][examples]. These examples also act as [integration test
 
 ## Python integration
 
-It is possible to use `rsmpi` with a communicator provided by mpi4py. An example project demonstrating this is available at
+It is possible to use `rsmpi` with a communicator provided by `mpi4py`. An example project demonstrating this is available at
 
 [https://github.com/betckegroup/mpi4py_with_rsmpi](https://github.com/betckegroup/mpi4py_with_rsmpi).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -213,6 +213,12 @@ See files in [examples/][examples]. These examples also act as [integration test
 
 [examples]: https://github.com/rsmpi/rsmpi/tree/master/examples
 
+## Python integration
+
+It is possible to use `rsmpi` with a communicator provided by `mpi4py`. An example project demonstrating this is available at
+
+[https://github.com/betckegroup/mpi4py_with_rsmpi](https://github.com/betckegroup/mpi4py_with_rsmpi).
```
jedbrown, line 220:
```suggestion
It is possible to use `rsmpi` with a communicator provided by [mpi4py](https://mpi4py.readthedocs.io/en/stable/). An example project demonstrating this is available at [mpi4py_with_rsmpi](https://github.com/betckegroup/mpi4py_with_rsmpi).
```

jedbrown (approved):
Looks good. https://github.com/betckegroup/mpi4py_with_rsmpi/pull/1

PR #131, opened 2022-08-25
Initial error handling

jtronge:
This is currently a WIP, but I wanted to open it now to get feedback since these would be breaking changes for any code that relies on rsmpi. I have not attempted anything yet with the collective code, but I've made all the code in `point_to_point.rs` return `Result`, indicating the MPI error class on failure. The examples should be updated as well.


src/error.rs
```diff
@@ -0,0 +1,258 @@
...
+        #[err(MPI_ERR_LASTCODE)]
+        LastCode,
+    }
+}
+
+/// Crate-internal function for mapping from MPI return code to RSMPI ErrorKind
+pub(crate) fn error_kind(res: c_int) -> ErrorKind {
+    match ErrorKind::from_raw(res) {
+        Some(kind) => kind,
+        None => panic!("Could not find matching ErrorKind for returncode '{}'", res),
```
jedbrown, line 256: Should we be calling `MPI_Error_string()` to be more descriptive?

src/topology/mod.rs
```diff
@@ -72,8 +73,17 @@ impl SystemCommunicator {
     ///
     /// # Examples
     /// See `examples/simple.rs`
-    pub fn world() -> SystemCommunicator {
-        unsafe { SystemCommunicator::from_raw_unchecked(ffi::RSMPI_COMM_WORLD) }
+    pub fn world() -> Result<SystemCommunicator, ErrorKind> {
+        unsafe {
+            let res = ffi::MPI_Comm_set_errhandler(ffi::RSMPI_COMM_WORLD, ffi::RSMPI_ERRORS_RETURN);
```
jedbrown, line 78: I don't think `world()` should imply this change of error handling behavior, and I don't think `world()` needs to return a `Result`.

jedbrown (commented):
Looks good in general; thanks! I think the razor is to ask whether a given function can fail for a reason other than programmer error. So `.world()` and `.this_process()` don't have errors. Certainly anything collective should return `Result` because it can fail due to network partition or disruption of other ranks; indeed these are errors some applications would want to actually recover from versus mere clean-up-and-exit.

Rust std has decided that allocation doesn't return `Result` (just panics). I can't envision a scenario in which rsmpi would be used in nostd so I think mere allocation need not return `Result`. So this raises the question of whether the likes of `.immediate_send()` should return `Result`. Perhaps not if it can't fail (when given valid inputs) except by something akin to failed allocation. (When we can't enforce "valid inputs" via the type system, we can panic if they are invalid, just like array indexing with an out of range index.)

I'd be curious to hear your and @hppritcha opinions about when to return `Result` and when to panic.

src/error.rs
hppritcha: yes that would be nice to do

src/topology/mod.rs
jtronge: Ok. Do you think we should just panic here on an error with `MPI_Comm_set_errhandler`? Or maybe I should move this call into `Universe::initialize()`.

jtronge:
Thanks for look at this.

For `immediate_send`/`MPI_Isend` and similar functions I'm not sure if there is a guarantee on what kinds of errors an implementation might return. It might be the case that it could return some unexpected but recoverable error, and in that case we probably don't want to panic. If it does only return allocation-related errors, then maybe it would be fine to panic there. But again I'm not sure about that.

It seems that MPI, when set to ERRORS_RETURN, will even return argument errors such as `MPI_ERR_ARG`, whereas in Rust code there would likely be an assert statement. I'm guessing that those types of errors would most likely point to an error in how RSMPI is validating user-provided arguments, but maybe not always.

I'm thinking that we should probably panic on some of these MPI errors, while returning the others. For example errors that might indicate that RSMPI code is doing something wrong or errors like `MPI_ERR_INTERN` which really isn't recoverable.

PR #140, opened 2022-12-05
Attributes, inter-communicators, and spawning

jedbrown:
This isn't super full-featured, but I think the interface is convenient enough. Process spawning with Open MPI is a bit more expensive than launching using `mpiexec`.


jtronge:
This looks great. I tried the three examples locally.

Is this assuming that inter communicators are created outside of rsmpi right now? I didn't see any `MPI_Intercomm_*` functions.

jedbrown:
`MPI_Comm_spawn` creates intercommunicators. I didn't bind `MPI_Intercomm_create` because I'm interested in spawning, but it could be done later.

Cargo.toml
```diff
@@ -27,14 +27,15 @@ derive = ["mpi-derive"]
 
 [dependencies]
 conv = "0.3"
-libffi = { version = "3.0.0", optional = true }
+libffi = { version = "3.0.1", optional = true }
```
hppritcha, line 30: were these version changes needed for this PR or just to pull in newer versions?

jedbrown: Not necessary; could be dropped.

PR #151, opened 2023-02-09
Fix parsing of MPI compiler flags.

### LINKED ISSUE #150: Incorrect parsing of `-L` and `-I` paths in build-probe-mpi
Some MPI implementations (namely Intel MPI) surround the linker and include paths in the output of `mpicc -show` with quotes, e.g.:

```sh
gcc -I"/opt/intel/oneapi/mpi/2021.8.0/include" -L"/opt/intel/oneapi/mpi/2021.8.0/lib/release" -L"/opt/intel/oneapi/mpi/2021.8.0/lib" -Xlinker --enable-new-dtags -Xlinker -rpath -Xlinker "/opt/intel/oneapi/mpi/2021.8.0/lib/release" -Xlinker -rpath -Xlinker "/opt/intel/oneapi/mpi/2021.8.0/lib" -lmpifort -lmpi -lrt -lpthread -Wl,-z,now -Wl,-z,relro -Wl,-z,noexecstack -Xlinker --enable-new-dtags -ldl
```

The current implementation of `build-probe-mpi` passes the substring after the `-I` and `-L` flags directly to `PathBuf::from`, which results in the string being parsed as a single file name and not as a path.

In the case of Intel MPI begin installed alongside OpenMPI, this causes the linkage of OpenMPI which does not result in a build failure, but leads to segfaults on program execution because some external symbols are incorrectly resolved.

I will add a PR which unescapes the paths from the `mpicc -show` output.

niklas-uhl:
This fixes a bug where the linker and include flags are not passed
correctly if the output of `mpicc -show` surrounds them by quotes.

Closes #150


jedbrown:
Thanks!

jedbrown:
Oops, can you use `cargo fmt` to fix the lint?

jedbrown:
Hi, could you rebase this on current `main` to pick up the CI versioning fix?

jedbrown:
Do you see what's going on with the tests?

niklas-uhl:
It seems to work now, see https://github.com/niklas-uhl/rsmpi/pull/1

jedbrown:
Thanks for fixing this issue and writing the tests. Is it possible we would get output with `"-L/path/with spaces/lib"`? Should we use [`shell_words::split`](https://docs.rs/shell-words/latest/shell_words/fn.split.html) for this?

niklas-uhl:
I tried installing Intel MPI to a location containing a space (was not possible with MPICH) and the output of `mpicc -show` was:
```bash
gcc -I"/opt/intel/My Oneapi/mpi/2021.8.0/include" -L"/opt/intel/My Oneapi/mpi/2021.8.0/lib/release" -L"/opt/intel/My Oneapi/mpi/2021.8.0/lib" -Xlinker --enable-new-dtags -Xlinker -rpath -Xlinker "/opt/intel/My Oneapi/mpi/2021.8.0/lib/release" -Xlinker -rpath -Xlinker "/opt/intel/My Oneapi/mpi/2021.8.0/lib" -lmpifort -lmpi -lrt -lpthread -Wl,-z,now -Wl,-z,relro -Wl,-z,noexecstack -Xlinker --enable-new-dtags -ldl
```

So yeah, the current parsing would fail in that case, which could be fixed by using `shell_words::split`.

But I have the impression that the whole argument parsing is rather fragile (even `mpicc -show` is not standardized). But being able to correctly link MPI programs is a crucial part for using MPI in Rust (e.g. I am actually not using the rsmpi bindings, but only the mpi-sys crate directly). Maybe one would be better of just calling a CMake script for fetching the MPI flags, because CMake's compile argument parsing is better maintained and mature. But this beyond the scope of this PR, maybe I will have a look at this when I have more time.

For getting this PR merged, do you still want me to include `shell_words::split` or make it a separate PR?

jedbrown:
Thanks. It's hard for me to imagine something I would dislike more than growing a dependency on CMake. :sweat_smile: 

Your updated PR looks great, thanks.

niklas-uhl:
Mhmm, the windows action succeeded before and I didn't touch any code that this depends on ...

niklas-uhl:
Ah, I saw that you already fixed a race condition in the examples, seems like there is another one.

jedbrown:
This one I've only ever seen on Windows, which I don't have on a dev machine. I actually don't know if it's a code problem or environment problem. I've seen Windows runners sometimes be flaky at earlier stages like git clone and I have much less confidence in correctness of MS-MPI than in MPICH or Open MPI.

Issue #148, opened 2023-01-17
Using dynamic reference to communicator


tbetcke:
Hi,

I'd like to store a communicator as dynamic reference, e.g.

```
let comm = &dyn Communicator;
```
This is so that users can provide any communicator that implements the ```Communicator``` trait without having to introduce a generic for it.

My problem is with using it. If I have a ```SystemCommunicator``` I can easily access communication methods (e.g. ```all_reduce_into```).

But it is not possible if I only have a reference to a ``` Communicator``` trait object.

Is there a way to use communication routines through a generic ```Communicator``` trait object? Or do I always require a concrete communicator? In the latter case how can we write code that supports different types of communicators?

jedbrown:
This part wasn't my design. I haven't thought about this carefully, but surely need this capability to write higher level libraries. Can you make a test and propose an implementation?

tbetcke:
Currently going trough the source code. Goal is to access ```all_reduce_into``` from a ```&dyn Communicator``` trait object.

```all_reduce_into``` is defined in the ```CommunicatorCollectives``` trait. This trait derives from ```Communicator``` and is implemented for any Communicator type. So ideally we would just cast 

```
(&dyn Communicator as &dyn CollectiveCommunicator).all_reduce_into(...)
```

However, this does not work because ```CollectiveCommunicator``` has template methods and cannot be cast into a trait object.

What seems to work is to implement ```CollectiveCommunicator``` for ```dyn Communicator``` that is to add in ```collective.rs``` the line

```imp CollectiveCommunicator for dyn Communicator```.

We would have to do this for all derived traits from Communicator in order for users to be able to use ```Communicator``` as trait object. I don't know if there are any other boundary effects from doing this. But this might be a solution.

tbetcke:
Looked more into the above. An issue is the ```process_at_rank``` method in the ```Communicator``` trait. It requires ```Self``` to be sized therefore cannot be called on the trait object.

jedbrown:
Surely that can be resolved with a little indirection at most.

tbetcke:
Agree. We would need to implement some stuff for ```dyn Communicator``` if we want it to have the same functionality as the direct communicator types. For example the ```process_at_rank``` method returns a new ```Process``` struct that does nothing else than storing a reference to the communicator + the rank information given to ```process_at_rank```.

The actual point to point communication is then implemented as a trait that only needs the rank information and the ```as_raw``` method from the communicator. I suspect the same with others.

[...]

Issue #32, opened 2018-02-06
Publish `is_initialized`


AndrewGaspar:
A scenario for rsmpi, which I imagine would be common, is to use it inside a static lib that is linked into a larger application. This larger application is likely written in another language like Fortran, C, or C++. The larger application is probably already initializing MPI with its preferred parameters. In this case, the Rust library could would not "own" the MPI initialization, and therefore would not call `mpi::initialize*`. However, it may still want to check if MPI has already been initialized, even if it is just to panic and tell the user that MPI needs to be initialized first.

Of course, `ffi::MPI_Initialized` is available, but it seems like making the Rust-native version of it available is pretty low-cost.

bsteinb:
Yeah, when I started on rsmpi I structured it in a way that is more geared toward using it from a rust application and having rsmpi be the only interface through which MPI is used. I later realized that writing a rust library which can then be used from HPC applications written in the more traditional languages is also a likely scenario.

However, there will have to be some restructuring to get rsmpi into shape for that use case. This issue touches on one aspect of it.

The `Universe` type will have to take a different role, maybe the initialized environment that it stands for will become fully implicit and not have value representation just as it is in the official MPI bindings. There could still be something like a `FinalizeGuard` in cases where the user wants something that calls `MPI_Finalize` on `drop`.

Types like `...Communicator` will need a `from_raw` constructor which check certain attributes of the underlying objects (intra-communicator vs. inter-communicator or persistent vs. non-persistent request).

Then of course objects like communicators that are passed in from another language could have different error handlers set. Right now, rsmpi assumes that communication operations crash on error, but that is not necessarily the case, so, in general, everything has to return `Result` just like (almost) everything returns `int` in the C bindings.

Issue #12 is also somewhat related to this.

If you don't mind, I will hijack this issue and make the title reflect the somewhat broader scope. It could then become a tracking issue for a few smaller issues that cover the different areas that need to be touched.

AndrewGaspar:
Ah, I hadn't considered all of the potential issues here. If you do find a way to add Send + Sync traits to MPI objects statically depending on while MPI threading setting used, you may need an `unsafe` mechanism for choosing at runtime whether Universe is safe to use in a multi-threaded environment. The caller would be expected to know how MPI was initialized in their context.

jedbrown:
[...]

PR #155, opened 2023-03-14
Remove type-distinction between system-communicators and user-communicators

Cydhra:
As outlined in #148, it is very unintuitive to split the Communicator interface in `SystemCommunicator` and `UserCommunicator`.
I drafted a PR to unify both into one common `enum SimpleCommunicator`:

- I migrated topology-related code from UserCommunicator into the common enum, so both communicator variants can now call those functions. I couldn't confirm it outright, but the spec does not prohibit that explicitely, so I think this models the MPI interface more closely.
- I refactored CartesianCommunicator to accept both World- and UserCommunicator, which is legal by spec, I don't know why this wasn't possible before.
- since WorldCommunicator is no longer Copy, code that moved a reference to SystemCommunicator must now explicitly refer to IntraCommunicator::WorldCommunicator (see immediate_multiple_test.rs)
- I added SelfCommunicator next to WorldCommunicator, because the MPI spec defines it and without it downstream crates would need unsafe code to create it, and call `free` on it when their `UserCommunicator` drops

Things missing from this draft:

There is another weird split in the code-base: InterCommunicators use `SystemCommunicatorHandle` and `UserCommunicatorHandle`, even though those handle structs do the exact same as their communicator-counterparts (at least as far as I can tell).
Since I have never worked with InterCommunicators before though, and there are neither integration-tests nor examples for them, I didn't migrate them to use `SimpleCommunicator` yet, as I wanted to get some feedback on their purpose.

All examples and tests are green, but I am not very familiar with the codebase yet, so somebody should scrutinize this.

### DIFF
--- a/examples/immediate_multiple_test.rs
+++ b/examples/immediate_multiple_test.rs
@@ -7,8 +7,8 @@ use mpi::Rank;
 const COUNT: usize = 128;
 
 /// Send and receive COUNT number of immediate requests.
-fn send_recv<'a, C: Communicator, S: Scope<'a> + Copy>(
-    world: C,
+fn send_recv<'a, S: Scope<'a> + Copy>(
+    world: &impl Communicator,
     scope: S,
     coll: &mut RequestCollection<'a, [i32; 4]>,
     next_proc: Rank,
@@ -54,15 +54,7 @@ fn main() {
         .collect();
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(
-            universe.world(),
-            scope,
-            coll,
-            next_proc,
-            prev_proc,
-            &x,
-            &mut recv,
-        );
+        send_recv(&world, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut buf = vec![];
         while coll.incomplete() > 0 {
@@ -75,15 +67,7 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(
-            universe.world(),
-            scope,
-            coll,
-            next_proc,
-            prev_proc,
-            &x,
-            &mut recv,
-        );
+        send_recv(&world, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut complete = vec![];
         let mut buf = vec![];
@@ -99,15 +83,7 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(
-            universe.world(),
-            scope,
-            coll,
-            next_proc,
-            prev_proc,
-            &x,
-            &mut recv,
-        );
+        send_recv(&world, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut complete = vec![];
         while !coll.test_all(&mut complete) {}
--- a/examples/immediate_multiple_test.rs
+++ b/examples/immediate_multiple_test.rs
@@ -1,9 +1,9 @@
 /// Example showing usage of test_any(), test_some() and test_all().
 use mpi;
 use mpi::request::{RequestCollection, Scope};
+use mpi::topology::SimpleCommunicator;
 use mpi::traits::*;
 use mpi::Rank;
-use mpi::topology::SimpleCommunicator;
 
 const COUNT: usize = 128;
 
@@ -55,7 +55,15 @@ fn main() {
         .collect();
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(
+            SimpleCommunicator::WorldCommunicator,
+            scope,
+            coll,
+            next_proc,
+            prev_proc,
+            &x,
+            &mut recv,
+        );
 
         let mut buf = vec![];
         while coll.incomplete() > 0 {
@@ -68,7 +76,15 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(
+            SimpleCommunicator::WorldCommunicator,
+            scope,
+            coll,
+            next_proc,
+            prev_proc,
+            &x,
+            &mut recv,
+        );
 
         let mut complete = vec![];
         let mut buf = vec![];
@@ -84,7 +100,15 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(
+            SimpleCommunicator::WorldCommunicator,
+            scope,
+            coll,
+            next_proc,
+            prev_proc,
+            &x,
+            &mut recv,
+        );
 
         let mut complete = vec![];
         while !coll.test_all(&mut complete) {}
--- a/examples/immediate_multiple_test.rs
+++ b/examples/immediate_multiple_test.rs
@@ -3,6 +3,7 @@ use mpi;
 use mpi::request::{RequestCollection, Scope};
 use mpi::traits::*;
 use mpi::Rank;
+use mpi::topology::SimpleCommunicator;
 
 const COUNT: usize = 128;
 
@@ -54,7 +55,7 @@ fn main() {
         .collect();
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(world, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut buf = vec![];
         while coll.incomplete() > 0 {
@@ -67,7 +68,7 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(world, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut complete = vec![];
         let mut buf = vec![];
@@ -83,7 +84,7 @@ fn main() {
 
     let mut recv: Vec<[i32; 4]> = vec![[0, 0, 0, 0]; COUNT];
     mpi::request::multiple_scope(2 * COUNT, |scope, coll| {
-        send_recv(world, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(SimpleCommunicator::WorldCommunicator, scope, coll, next_proc, prev_proc, &x, &mut recv);
 
         let mut complete = vec![];
         while !coll.test_all(&mut complete) {}
--- a/examples/immediate_multiple_test.rs
+++ b/examples/immediate_multiple_test.rs
@@ -1,7 +1,6 @@
 /// Example showing usage of test_any(), test_some() and test_all().
 use mpi;
 use mpi::request::{RequestCollection, Scope};
-use mpi::topology::SimpleCommunicator;
 use mpi::traits::*;
 use mpi::Rank;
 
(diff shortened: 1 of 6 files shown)


jedbrown:
Thanks! The `spawn` examples use inter-communicators. A subtle point is that the inter-comm returned by `MPI_Comm_get_parent` should be thought of as a "system" inter-comm since freeing it would leave other references dangling. This is in contrast to the inter-comm returned by `MPI_Comm_spawn`, which has normal "user" semantics. This might help explain why the separate handles exist. I'm not attached to this design, BTW.

```rs
/// A system-defined inter-communicator
pub type SystemInterCommunicator = InterCommunicator<SystemCommunicatorHandle>;

/// A user-defined inter-communicator
pub type UserInterCommunicator = InterCommunicator<UserCommunicatorHandle>;
```

Looks like we have some CI failures.

Cydhra:
> A subtle point is that the inter-comm returned by `MPI_Comm_get_parent` should be thought of as a "system" inter-comm since freeing it would leave other references dangling. This is in contrast to the inter-comm returned by `MPI_Comm_spawn`, which has normal "user" semantics. This might help explain why the separate handles exist. I'm not attached to this design, BTW.

The reason why the separate handles exist is the same why `SystemCommunicator` and `UserCommunicator` were split in the first place, I suppose. My question was why use a separate -`Handle` instead of `SystemCommunicator` and `UserCommunicator`.

I don't quite understand if Drop-semantics are important here: Intuitively I wouldn't even unify the two handles into a common `SimpleCommunicatorHandle`, but instead use `SimpleCommunicator` directly. But I don't know if there was a reason to differentiate between `*CommunicatorHandle` and `*Communicator`.

> Looks like we have some CI failures.

I was bitten by `#[cfg(not(msmpi))]`, since I tested everything on Windows. Fixing it on Fedora right now.

src/topology/mod.rs
```diff
@@ -246,74 +148,123 @@ impl UserCommunicator {
...
 }
 
-impl From<CartesianCommunicator> for UserCommunicator {
-    fn from(cart_comm: CartesianCommunicator) -> Self {
-        cart_comm.0
-    }
-}
+/// A handle to a communicator used in InterCommunicators. This type is functionally similar to `SimpleCommunicator`
+/// save for its Drop-semantics.
+pub enum SimpleCommunicatorHandle {
```
jedbrown, line 177: I think we have a weird middle ground here where there is a type-level distinction between system and user inter-communicators, but not for intra-communicators. `Self` and `World` are not inter-communicators.

`MPI_Com_get_parent` returns a "system" inter-comm, with possibly surprising semantics.

image

I think we could make `InterCommunicator` an `enum` and remove `*Handle`s entirely. Or we could have more privacy by making a single `enum CommunicatorHandle` and having a non-public (perhaps `pub(crate) CommunicatorHandle`) field in `SimpleCommunicator`, `IntraCommunicator`, and the topology communicators. So all handling of a raw `MPI_Comm` would be inside `CommunicatorHandle` and yet we wouldn't be revealing those publicly. This would prevent users from pattern matching on whether a given communicator is `World` (for example), but that's a code smell anyway, and it would prevent future internal changes from breaking user code. What do you think?

examples/immediate_multiple_test.rs
```diff
@@ -67,7 +76,15 @@ fn main() {
...
-        send_recv(world, scope, coll, next_proc, prev_proc, &x, &mut recv);
+        send_recv(
+            SimpleCommunicator::WorldCommunicator,
+            scope,
+            coll,
+            next_proc,
+            prev_proc,
+            &x,
+            &mut recv,
+        );
```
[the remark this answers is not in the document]
Cydhra: Oh yeah that is probably the better solution for this particular case. I wasn't thinking this far because when I saw that the previous code broke, I worried about the *double free* problem.

src/topology/mod.rs
Cydhra: The reason `World` is still in this enum is that I assume `MPI_COMM_GET_PARENT` can return WORLD as a parent. From the spec, it sounds like this cannot happen, but the codebase explicitely defined an InterComm with `SystemCommunicatorHandle`, so I assumed that it is necessary. 

If it is, then making the `Handle` an `enum` with `pub(crate)` fields in the relevant structs is probably the most elegant solution. We would remove the `Drop` implementation from the `Handle` and instead add it to the public wrapper types `SimpleCommunicator`, `InterCommunicator` and the topology communicators.

If all of those communicators implement `Communicator` and this trait then contains all methods like it was initially suggested in #148, the split in the type system is gone completely. My solution so far only removes it between `System*` and `User*`, not between intra- and intercommunicators.

jedbrown: `MPI_Comm_get_parent` cannot return `WORLD`. We had `InterCommunicator<SystemCommunicatorHandle>` so that `world.parent()` can be called multiple times without returning None or leaving dangling references.

I was thinking that `Handle` would implement dropping semantics so that code doesn't clutter the higher levels (public types). I think `SimpleCommunicator`, `InterCommunicator`, and the topological communicators would be publicly visible types because they have some arguments that don't apply to general communicators. I haven't thought about whether we need an `IntraCommunicator` trait (implemented by `SimpleCommunicator` and the topological comms) so it can be used as a type constraint.

Cydhra: If the `Handle` is used in both Inter- and Intracomms, Drop cannot be implemented on the handle, because it must either call `MPI_COMM_DISCONNECT` or `MPI_COMM_FREE`.

jedbrown: If `Handle` is an enum, its `drop` implementation can match on whether to do nothing (World, Self, and parent), to `MPI_Comm_disconnect` as with a child inter-comm, or to `MPI_Comm_free` (user intra-comm). The constructors would need to create the correct handle before putting it in a public type, but then nothing else would need to implement `drop`.

Cydhra: Sorry for the delay, I am amidst exam season. I will  try to implement the second option, turning the Handle into an enum and hiding it from the public API. As you were saying, this reduces code smell by smoothing the distinction between World- and User-comms, reduces the overall API complexity, and hides away the raw handles properly. This doesn't address the initial concern of a unified `Communicator` trait yet, but as suggested in the issue we could then probably move everything into `trait Communicator` and use `dyn Communicator` for downstream crates.

The only problem I currently see are immediate requests, because they consume the communicator, which may introduce new complexity once the raw communicator is hidden away. I haven't looked into that yet, I assume this has to do with RAII semantics. I will ignore it for now, but this issue needs to be looked into and fixed before merging.

Cydhra:
I have implemented an enum `CommunicatorHandle` now, deleted the previous handles and their trait, simplified the API to one struct `SimpleCommunicator`, one `InterCommunicator` and the topological communicators.

The latter still use simple communicators in their constructor, I don't know if that is how it is supposed to be, because it seems weird to have this conversion going on. I need some feedback on that.

The Handle type is still public, I don't know if we should hide it completely. At the moment hiding it is weird, because there is the public trait `AsRaw` that exposes the raw handles from ffi anyway. If we want to hide the handle enum, we should also hide `AsRaw`.

jedbrown:
Cool, I'll check it out as time allows.

I think of `AsRaw` as public because a user may want to call a C function (say PETSc) that takes an `MPI_Comm` (relates to #32). My current thinking is that `AsRaw` can stay public while hiding the handle. One should be able to create any of the concrete implementations from a raw `MPI_Comm`, but I don't think we need to see the `CommunicatorHandle` for that either. Do you have a use case?

Cydhra:
> Cool, I'll check it out as time allows.
> 
> I think of `AsRaw` as public because a user may want to call a C function (say PETSc) that takes an `MPI_Comm` (relates to #32). My current thinking is that `AsRaw` can stay public while hiding the handle. One should be able to create any of the concrete implementations from a raw `MPI_Comm`, but I don't think we need to see the `CommunicatorHandle` for that either. Do you have a use case?

Nope. At the moment `InterCommunicator` can be created from `CommunicatorHandle` (which is a bit weird) so this needs to be changed to hide the handle. I wanted to ask whether to hide the handle before changing it, because it was like this before. For some reason the check whether a raw handle is of an inter-comm was implemented in the InterComm constructor instead of the Handle-constructor. I have since changed that.

src/topology/cartesian.rs
```diff
@@ -51,8 +52,8 @@ impl CartesianCommunicator {
     /// # Safety
     /// - `raw` must be a live MPI_Comm object.
     /// - `raw` must not be used after calling `from_raw`.
-    pub unsafe fn from_raw(raw: MPI_Comm) -> Option<CartesianCommunicator> {
-        UserCommunicator::from_raw(raw).and_then(|comm| match comm.into_topology() {
+    pub unsafe fn from_raw_checked(raw: MPI_Comm) -> Option<CartesianCommunicator> {
```
jedbrown, line 55: Why did `from_raw` change from returning an `Option` to an unchecked version? Can we avoid that without contortions? In general, I'd rather discourage using unchecked versions.

jedbrown (commented):
Thanks for your work here and sorry for my slow review. Would it be a lot to ask for an example showing something that can be done in the new model that couldn't be done before?

I suspect the new version can compile a little faster/smaller due to fewer variants to monomorphize, though even that is hard to measure. I find it a bit more intuitive, but it seems to be almost exactly the same amount of code. Unfortunately, we can still hardly use `&dyn Communicator` because most methods require `Self: Sized`. I'm not sure how to gracefully get us out of that bind.

src/topology/mod.rs
```diff
@@ -54,159 +54,213 @@ pub trait AsCommunicator {
...
 
-unsafe impl AsRaw for SystemCommunicator {
-    type Raw = MPI_Comm;
-    fn as_raw(&self) -> Self::Raw {
-        self.0
-    }
-}
+    /// Create a `CommunicatorHandle::InterCommunicator` from a raw handle. Returns `None` if the
+    /// handle is `MPI_COMM_NULL`.
+    pub fn inter_from_raw(raw: MPI_Comm) -> Option<CommunicatorHandle> {
```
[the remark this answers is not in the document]
Cydhra: Probably? The use of `unsafe` in the API does not follow a pattern I could recognize. I didn't use unsafe when the function performed checks and returned an `Option` because that is largely how it was implemented before my changes.

src/topology/mod.rs
```diff
@@ -54,159 +54,213 @@ pub trait AsCommunicator {
...
-    /// - `raw` must not be `MPI_COMM_NULL`.
-    unsafe fn from_raw_unchecked(raw: MPI_Comm) -> UserCommunicator {
+    /// Do not call with inter-comm handles and inter-comm parent handles, `MPI_COMM_WORLD`, or
+    /// `MPI_COMM_SELF`
+    unsafe fn from_raw_unchecked(raw: MPI_Comm) -> SimpleCommunicator {
         debug_assert_ne!(raw, ffi::RSMPI_COMM_NULL);
-        UserCommunicator(raw)
+        let handle = CommunicatorHandle::simple_from_raw_unchecked(raw);
+        SimpleCommunicator(handle)
     }
```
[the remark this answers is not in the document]
Cydhra: I agree. The problem is, I tried to minimize the havoc I caused: I initially started this PR, because I wanted to unify the communicators, which would side-step the `&dyn Communicator` problem. I didn't realize that inter-communicators behaved differently, so this endeavor failed.

I continued to at least clean up the code-base by unifying the raw handle like you suggested, but I didn't want to blindly remove functionality that was there before. I think we shouldn't expose the unchecked constructors of any structure to public API, but I tried not to disrupt the API too much. I will gladly go over it again though, to check which unchecked constructors can be removed, and what checks can be added to the remaining constructors, if that is desired.

To this end, there is another question though: `FromRaw` is a trait, for which I retrofitted implementations into the communicator and handle structs, because you wanted downstream crates to be able to call them with raw handles (and it seemed stupid to not use the already existing trait, even if it wasnt implemented before). I used unsafe and unchecked implementations for those, because the trait does not return an `Option` and was already marked `unsafe`. Should this trait and its implementations remain, and/or should there be a safe alternative trait?

src/topology/cartesian.rs
Cydhra: Because I implemented `FromRaw` for everything and that already used the name `from_raw` for the unchecked version, so I renamed the checked version of the function.

Cydhra:
About the example what can be done in the new model:
I don't think it changes that much to be honest. I initially wanted to change it, because I was using `&SystemCommunicator` in my crate, which was overly restrictive. So I tried to unify it with `UserCommunicator` (because as #148 points out, `dyn Communicator` doesnt work), but then realized that I cannot easily unify the struct with `InterCommunicator`.

I then changed my approach to your suggestion of at least encapsulating the handle correctly, which further disconnected this PR from its initial attempt to make it easier to support different communicators for higher-level routines: Now the distinction between Intra-, Inter- and topological communicators still exists publicly, only the handle is nicely modelled. I am still hoping this will simplify the solution for #148, but from the outside, this PR doesn't improve the API that much, sadly.

About all the safety-related stuff: I tried to not create more mess than was already present, but some functions were and are not marked unsafe that really should be (then again, at least some collective communication functions are unsound but not marked unsafe either).

src/topology/cartesian.rs
Cydhra: If you are worried about breaking changes: Since we remove types like `SystemCommunicator` we are already breaking. And this won't silently change semantics since the return-type changed.

Would a mention of the checked version in the docs suffice, or should we hide the unchecked conversions from public API entirely, or should I revert those changes and declare it out-of-scope for this PR?

jedbrown: I'm not that worried about the breaking change (so long as it's easy for users to update), but rather consistency and encouraging safer use. When in doubt, people usually choose the shorter version of the function. Looking at core, it seems the patterns could be `from_raw, try_from_raw` or `from_raw_unchecked, from_raw`. I guess either is okay with me. I didn't see much prior art for the `from_raw_checked` naming, but maybe you have.

src/topology/mod.rs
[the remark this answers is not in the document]
Cydhra: Okay so, I implemented the `FromRaw` trait for nearly everything (because you wanted a way for users to create the types from raw handles) and also tried to provide `try_from_raw` versions everywhere.

Should I get rid of unchecked versions (implemented via the `FromRaw` trait) where we don't need them, or is it fine the way it is (it mostly was like this before).

Cydhra:
Is this still desired?

jedbrown:
Sorry, I got swamped with end-of-semester stuff. I'll be able to put more attention toward this next week.

jedbrown (commented):
Thanks. Sorry about losing track of this. Pipeline is running now.

PR #159, opened 2023-04-17
Implement Equivalence trait for complex types

### LINKED ISSUE #149: Implement Equivalence trait for complex types
Hi,

the ```Equivalence``` trait is currently not implemented for complex types (see types in ```datatype.rs```).
It would be simple to add  corresponding definitions for the types ```c32``` and ```c64``` defined in the ```Cauchy``` crate.
This might make sense as this is a widely used package for complex types. However, it would introduce a dependency to ```Cauchy```. Is this possible or not a preferable option for ```rsmpi```? If possible, I will be happy to write a pull request for it.

keefehuang:
Closes #149.

Adds feature "complex" to top level Cargo.toml with dependency on num_complex library.
Externs MPI_C_FLOAT_COMPLEX and MPI_C_DOUBLE_COMPLEX.
Implements Equivalence trait for Complex types from `num_complex` to C MPI implementations by directly mapping to MPI_C_FLOAT_COMPLEX and MPI_C_DOUBLE_COMPLEX.


jedbrown:
Thanks! Could you add a test? I don't think it needs to go in `examples/`, but either `tests/` or a unit test in `datatype.rs`?

jedbrown:
Hi @keefehuang, just checking if you'd be able to rebase on main (to settle the conflict above) and add a test for this new feature?

keefehuang:
Ah sorry. I've added unittests into the same file complex.rs. Which additional tests should I include?

Cargo.toml
```diff
@@ -24,14 +24,20 @@ default = ["user-operations"]
...
+
 
 [dependencies]
 conv = "0.3.3"
 libffi = { version = "3.2.0", optional = true }
 # Public dependency ("derive" feature)
-memoffset = {version = "0.9", optional = true }
+
+num-complex = { version = "0.4.3", optional = true }
+memoffset = { version = "0.9", optional = true }
```
jedbrown, line 36: I think the comment is intended to apply to `memoffset`. The comment could be deleted since it's redundant with the `[features]` block above.

Cargo.toml
```diff
@@ -24,14 +24,20 @@ default = ["user-operations"]
...
 mpi-derive = { path = "mpi-derive", version = "0.1", optional = true }
 mpi-sys = { path = "mpi-sys", version = "0.2" }
+
+# Public dependency ("derive" feature)
```
jedbrown, line 40: I think this comment is not quite accurate since it's needed regardless of the `derive` feature.

examples/complex_numbers.rs
```diff
@@ -0,0 +1,38 @@
+#![deny(warnings)]
+#![allow(forgetting_copy_types)]
+extern crate mpi;
+#[cfg(feature = "complex")]
+use mpi::traits::*;
+#[cfg(feature = "complex")]
+use num_complex::Complex64;
+
+fn main() {
+    #[cfg(feature = "complex")]
```
jedbrown, line 10: Please have a look at the `[[example]]` blocks in `Cargo.toml` for entire examples that require a feature.

jedbrown:
I think you can also use `--features complex,derive` for the windows runner (`.github/workflows/test.yaml`). We don't have `user-operations` on Windows so can't use `--all-features` there.

examples/complex_numbers.rs
keefehuang: Ok. Updated to add a new [[example]] block and removed unneeded #cfg blocks

Cargo.toml
keefehuang: Indeed. Removed both blocks and tidied up.

examples/complex_numbers.rs
```diff
@@ -0,0 +1,35 @@
...
+
+        root_process.broadcast_into(&mut data);
+
+        assert_eq!(
+            vec![
+                Complex64::new(1.0, 2.0),
+                Complex64::new(2.0, 3.0),
+                Complex64::new(3.0, 4.0),
+            ],
+            data
```
jedbrown, line 32: I'm getting this, suggesting there is a problem with the implementation.
```console
$ cargo mpirun -n 2 --features complex --example complex_numbers
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
thread 'main' panicked at examples/complex_numbers.rs:26:9:
assertion `left == right` failed
  left: [Complex { re: 1.0, im: 2.0 }, Complex { re: 2.0, im: 3.0 }, Complex { re: 3.0, im: 4.0 }]
 right: []
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

examples/complex_numbers.rs
```diff
@@ -0,0 +1,35 @@
+#![deny(warnings)]
+#![allow(forgetting_copy_types)]
```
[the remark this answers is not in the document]
keefehuang: Ok. These have been removed. Didn't prompt any errors or warnings during test, so this should work.

examples/complex_numbers.rs
keefehuang: Ah fixed. Shouldn't use ::capacity, simply initialized new zero entry complex vector. Also fixed.

PR #160, opened 2023-05-11
Add RSMPI_UNWEIGHTED

mscroggs:
This is required by https://github.com/linalg-rs/rlst/issues/20

Minimal example that fails with openmpi but runs with mpich on current rsmpi main: https://github.com/mscroggs/mpi-minfail


mscroggs:
Converted this to a draft as my failing example is now passing with openmpi but failing with mpich

mscroggs:
I've made RSMPI_UNWEIGHTED into a function as this was the only way I could find to make this work with both openmpi and mpich.

@jedbrown perhaps you have an idea for how to better do this?

jedbrown:
Can you share a minimal runnable test? The one in mpi-minfail doesn't run, and returns a communicator that is dropped after MPI_Finalize.

`MPI_UNWEIGHTED` is in the batch of constants that cannot be used in Fortran initialization expressions, but as I understand it, should be link-time constants in C. Looking at the MPICH implementation, it is initialized as a link-time constant, so there must be some interaction I don't yet understand. The function is okay if indeed necessary.
image

jedbrown:
Also note that support for graph communicators would ideally move into rsmpi itself, though I understand that may be a bigger ask.

mscroggs:
> Can you share a minimal runnable test?

I resorted to running cargo test in [rlst/sparse](https://github.com/linalg-rs/rlst/tree/main/sparse) to test this. I'll work on a more minimal test

mscroggs:
> Can you share a minimal runnable test?

I've added a minimal example to the mpi-minfail repo: https://github.com/mscroggs/mpi-minfail/blob/main/examples/test.rs

From the CI runs, you can see that this fails with openmpi using the current rsmpi (runs tagged "rsmpi"), but passes for both mpich and openmpi with this patch (runs tagged "mscroggs"): https://github.com/mscroggs/mpi-minfail/actions/runs/4979552904

jedbrown:
Oh, I understand the issue now and this solution is fine. The original values can be used as initializers inside functions, but not with static storage duration. People don't usually notice this because they're initializing local variables (inside functions, which have "automatic" storage duration).

Looks like we need to bump MSRV to 1.60 or pin a version on a dependency. Can't wait for this MSRV problem to be resolved by Cargo.

jedbrown:
Alright, please rebase on `main` and this should be good to go.

PR #171, opened 2023-11-25
Refactor attributes and add support for setting

jedbrown:
Attributes now hide the keys from most user code, and are easy to implement for new types. The current implementation still does not support setting/getting attributes on `dyn Communicator`.

### DIFF
--- a/src/topology/mod.rs
+++ b/src/topology/mod.rs
@@ -30,6 +30,7 @@ use conv::ConvUtil;
 use crate::Tag;
 use crate::{Count, IntArray};
 
+use crate::attribute::CommAttribute;
 use crate::datatype::traits::*;
 use crate::ffi;
 use crate::ffi::{MPI_Comm, MPI_Group};
@@ -40,7 +41,7 @@ mod cartesian;
 
 /// Topology traits
 pub mod traits {
-    pub use super::{AsCommunicator, Attribute, Communicator, Group};
+    pub use super::{AnyCommunicator, AsCommunicator, Communicator, Group};
 }
 
 // Re-export cartesian functions and types from topology modules.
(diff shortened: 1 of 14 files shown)


src/topology/mod.rs
```diff
@@ -843,6 +820,70 @@ pub trait Communicator: sealed::AsHandle {
     }
 }
 
+/// Methods that would otherwise block object safety.
+pub trait CommunicatorAux: Communicator {
```
jedbrown, line 824: @Cydhra This demonstrates how to handle attributes in a way that works with `&dyn Communicator`. I don't like the name `CommunicatorAux`, but not sure how to name it clearly. Ideally nobody would need to know about it because calling its methods are transparent due to the blanket impl (below), but it will show through in error messages. Suggestions welcome.

jedbrown: Basically everything currently in `Communicator` could move, and that would remove the remaining `Self: Sized` bounds. This is the only method that's being implemented.
```rs
impl Communicator for InterCommunicator {
    fn target_size(&self) -> Rank {
        self.remote_size()
    }
}
```

Cydhra: I agree that the name is a bit cumbersome, but I can't think of a better alternative either. The trait has no merit from an engineering perspective, it's just an artifact of a weird restriction in Rust. I guess, document why it exists and leave it as is?

Cydhra: No I get the technical rationale, I was talking from a language-agnostic engineering point of view. Anyway, yes, I don't see better alternatives than the names you suggested.

jedbrown: Thanks, I think I like `AnyCommunicator` given the other `*Communicator` names.

PR #172, opened 2023-12-16
docs: fix homepage example indentation

HoloTheDrunk:
Sorry about the microscopic PR but this has been bugging me for days


jedbrown (commented):
Thanks. How would you feel about making this a more systemic fix? See the github lint.yaml and change it to use
```
cargo fmt --all -- --config format_code_in_doc_comments=true --check
```
That should be fine since it always uses the nightly toolchain. We can't add it to `.rustfmt.toml` because it'll break ability to `cargo fmt` with a stable toolchain (until https://github.com/rust-lang/rustfmt/issues/3348 stabilizes).

HoloTheDrunk:
Added, sorry for the unreasonable delay - things got busy.

jedbrown:
I fixed some things up (applying the changes and some other CI volatility fixes from main) and merged. Thanks for your contribution.

Issue #174, opened 2024-01-21
Improving ergonomics of sending arrays/vectors


EdmundGoodman:
# Motivation

Currently, arrays and vectors of data cannot be represented by the `ReceiveFuture` type, as they do not implement the `Equivalence` trait. This means that functions like `immediate_receive` cannot be used with them.

I am working on porting a High-Performance Computing benchmark from C++ to Rust, which involves sending and receiving arrays or vectors of data. Looking through the documentation, there are various ways to do this, such as using `receive_vec` or `immediate_receive_into`. However, it would be useful if functions like `immediate_receive` also supported these types.

# Example

I would like to be able to write code like this:

```rust
use mpi::traits::*;

fn main() {
    let universe = mpi::initialize().unwrap();
    let world = universe.world();

    let to_send = [1, 2, 3];
    let future = world.any_process().immediate_receive::<[i32; 3usize]>();
    world.this_process().send(&to_send);
    let (msg, status) = future.get();

    println!("{:?} = {:?}", to_send, msg);
    println!("{:?}", status);
}
```

However, this fails as it requires `[T; D]` to implement the `Equivalence` trait.

# Attempted solution

I attempted to create a fix which I could PR to resolve this by adding an implementation to `datatype.rs`:

```rust
unsafe impl<T, const D: usize> Equivalence for [T; D]
where
    T: Equivalence,
{
    type Out = UserDatatype;

    fn equivalent_datatype() -> Self::Out {
        UserDatatype::contiguous(D as i32, &T::equivalent_datatype())
    }
}
```

Unfortunately, this results in conflicting implementations for all other traits on `[T; D]` in `datatype.rs` (`AsDatatype`, `Collection`, ...), and I am not entirely sure why all these other traits can co-exist, but `Equivalence` conflicts with all of them.

I am very happy to have a further look at this and PR to fix it, but would appreciate some advice on the direction to work in, as I am slightly stuck on where to go from here.

jtronge:
It looks like you might be able to fix the error by removing all `[T; D]` implementations for `AsDatatype`, `Collection`, etc. These are getting implemented twice since there's already a blanket implementation that does this. For example, with `Buffer` the trait is getting implemented twice by these lines:

* `unsafe impl<T> Buffer for T where T: Equivalence {}`
* `unsafe impl<T, const D: usize> Buffer for [T; D] where T: Equivalence {}`

Removing the second one should get rid of the error.

I'm not sure if this might break other things though.

[...]

PR #176, opened 2024-01-22
Add equivalence trait for arrays, with integration tests

EdmundGoodman:
This PR resolves the first part of #174, by adding the `Equivalence` trait to arrays, allowing them to be used without wrappers in places which require the trait, such as `ReceiveFuture` or `immediate_send`.

In order to do this, some other trait implementations in `datatype.rs` already covered by a blanket implementation have to be removed to avoid conflict. This does not impact `cargo test` cases passing. If there are significant worries that removing these implementations could change behaviours in strange edge cases, something I could change on review for it is adding it as optional functionality similar to `num-complex`.

Additionally, integration tests for this ergonomic new way of using arrays are added in `arrays.rs`, and checking their equivalence properties in `struct.rs`.

### DIFF
--- a/examples/arrays.rs
+++ b/examples/arrays.rs
@@ -1,0 +1,22 @@
+#![deny(warnings)]
+
+use mpi::traits::*;
+
+fn main() {
+    let universe = mpi::initialize().unwrap();
+    let world = universe.world();
+
+    let to_send = [1, 2, 3];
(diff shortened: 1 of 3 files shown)


EdmundGoodman:
Realised that `cargo test` does not run all the integration test examples - will run `run-examples.sh` and attempt to diagnose any issues that present themselves there.

PR #192, opened 2024-08-14
Update README URLs

Pencilcaseman:
One of the links was incorrect, and a couple referred to the old branch `master`, which has since been renamed `main`. These redirected correctly, but it was an easy fix :)


jedbrown (approved):
Thanks, this looks good.

Issue #189, opened 2024-07-16
Derive implementation of `Equivalence` hard codes `mpi` path


2pt0:
The `Equivalence` derive macro expects `mpi` to be an explicit dependency of the crate using the derive macro by hard coding the `mpi` dependency in the paths of the types used in the proc macro, e.g. `::mpi::some::field`. An error is returned if `mpi` is provided some other way, e.g. as a re-export in a dependency.

My project is using a workspace where most of the MPI code has been implemented in `crate1`. `crate1` re-exports `mpi`. `crate2` uses `crate1` as a dependency and as a way to access the `mpi` crate via `crate1::mpi`. Deriving `crate1::mpi::traits::Equivalence` in `crate2` a compile time error in this case because `mpi` is not a direct dependency of `crate2`.

Currently, my project can compile by making `mpi` a workspace dependency and including it as a dependency for `crate1` and `crate2`. However, if a user were to use my `crate1` as a dependency, they would not be able to derive `crate1::mpi::traits::Equivalence`.

jedbrown:
Is there an example or recommended strategy to handle this problem?

2pt0:
I have very little experience with proc macros, but after a little research I've found that `serde`, e.g., provides a [`crate` attribute](https://serde.rs/container-attrs.html#crate). Though I can't find it documented, `bytemuck` also provides a [`crate` attribute](https://github.com/Lokathor/bytemuck/blob/main/derive/src/traits.rs#L1313)

The solution would look end up looking like this:

```rust
use crate1::mpi::traits::Equivalence;

#[derive(Equivalence)]
#[mpi(crate = "crate1::mpi")]
struct Particle {
    position: [f64; 3]
}
```

jedbrown:
Thanks. This seems like an acceptable solution. I've added the https://github.com/rsmpi/rsmpi/labels/help%20wanted label since I can't work on it this week and I think it could be a good first contribution given the prior art.

PR #193, opened 2024-08-16
Allow custom crate paths for Equivalence derive macro

Cydhra:
The `Equivalence` derive macro cannot be used if the `mpi` crate is used from a re-exported path (#189). 
This PR fixes the issue with the suggestion by @2pt0:

```rs
use crate1::mpi_derive::Equivalence;

#[derive(Equivalence)]
#[mpi(crate = "::crate1::mpi")]
struct Particle {
    position: [f64; 3]
}
```

I wrote the new attribute parsing with neat little error messages, which the rest of the derive macro doesn't have (it just panics). Might be an idea for a future improvement.

Limitations:
- If the re-export doesn't enable the `derive` feature, the derive macro panics with very unhelpful error messages (missing modules). This cannot be addressed easily, because `cfg!(feature = "derive")` cannot be checked within the macro crate.
- This PR does not contain any test code because I don't know how to automatically test compilation stuff.
- I changed no documentation. This should probably be remedied before merging.

### DIFF
--- a/mpi-derive/src/lib.rs
+++ b/mpi-derive/src/lib.rs
@@ -17,7 +17,10 @@ pub fn create_user_datatype(input: TokenStream1) -> TokenStream1 {
     result.into()
 }
 
-fn equivalence_for_tuple_field(mpi_crate_path: &TokenStream2, type_tuple: &syn::TypeTuple) -> TokenStream2 {
+fn equivalence_for_tuple_field(
+    mpi_crate_path: &TokenStream2,
+    type_tuple: &syn::TypeTuple,
+) -> TokenStream2 {
     let field_blocklengths = type_tuple.elems.iter().map(|_| 1);
 
     let fields = type_tuple
@@ -26,7 +29,10 @@ fn equivalence_for_tuple_field(mpi_crate_path: &TokenStream2, type_tuple: &syn::
         .enumerate()
         .map(|(i, _)| syn::Index::from(i));
 
-    let field_datatypes = type_tuple.elems.iter().map(|ty| equivalence_for_type(mpi_crate_path, &ty));
+    let field_datatypes = type_tuple
+        .elems
+        .iter()
+        .map(|ty| equivalence_for_type(mpi_crate_path, ty));
 
     quote! {
         &#mpi_crate_path::datatype::UncommittedUserDatatype::structured(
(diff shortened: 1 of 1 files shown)


jedbrown (commented):
Thanks for implementing this, and sorry about my big delay reviewing it. The implementation makes sense to me after some background reading. Would it be possible to test that it works and mention it in docs? (I fear this functionality may be forgotten if someone refactors the macro in the future.)

Cydhra:
Mentioning it in the docs should be no problem. I don't know how to properly test this, though. Ideally, I'd just write an integration test that tries to compile something using this macro. However, if the macro is implemented incorrectly, the test would still compile, because I cannot remove `::mpi` as a dependency to the integration test.

```rs
#![cfg(feature = "derive")]

use mpi_derive::Equivalence;

// pretend the mpi crate is available at a custom path. We cannot disable mpi being a dependency
// for this test, so ::mpi is always available.
use mpi as custom_crate_path;

/// We test if the #[mpi(crate)] macro works correctly.
/// This should compile if the macro correctly uses custom_crate_path to address mpi.
/// Note that this also compiles if the macro does nothing, because we can't hide the mpi dependency
/// so this test can check whether the implementation of the macro is incorrect, but not if
/// it is empty.
#[test]
fn derive_custom_path() {
    #[derive(Equivalence)]
    #[mpi(crate = "custom_crate_path")]
    struct Particle {
        position: [f64; 3],
    }
}
```

jedbrown:
How about make a new crate `test-derive`, then use
```toml
[dependencies]
rsmpi = { package = "mpi", version = "0.8.0", path = ".." }
```
and in the Rust code,
```rs
use rsmpi::traits::Equivalence;
```
That won't work for me without using this branch and `#[mpi(crate = "rsmpi")]`. Would that sufficiently exercise what you need?

Cydhra:
This works. Keep in mind that `cargo test` doesn't run that test, so you will have to modify the GitHub runner. I'd rather not mess with it, since I have zero experience with them, and don't want to break it.

Cydhra:
Documentation is still missing, but tbh I don't know where the derive macro is documented?

jedbrown:
Oh, thanks. There is a brief blurb about `derive(Equivalence)` in `README.md`. I think that example could be copied to docs for the `Equivalence` trait since that's where someone would look. If you don't want to put a derive example in the trait documentation, it refer to [`mpi_derive::Equivalence`](https://docs.rs/mpi-derive/latest/mpi_derive/derive.Equivalence.html) (just write doc comments above `#[proc_macro_derive(Equivalence, attributes(mpi))]`).

Cydhra:
As far as I can tell there is no way to provide code examples to derive macros that don't fail compilation. Within the macro crate, the trait crate (mpi) cannot be included, and vice versa. So I disabled doctests for the examples and left a note in the macro-test case which I added yesterday to check them, if the test case needs to be modified.
Very unfortunate solution, but writing doc tests with multiple dependencies seems to be impossible.

jedbrown:
Wonderful, thanks! I appreciate the concern about testing these docs. Maybe rustdoc will support this in the future.

PR #195, opened 2024-08-20
Create `rustfmt.toml` and reformat accordingly -- Closes #194

### LINKED ISSUE #194: Create `rustfmt.toml`
There is no `rustfmt.toml` file in the project, so running `cargo fmt` has no well-defined style. For example, running `cargo fmt --check` on my machine complains an awful lot...

Based on the [CI linter script](https://github.com/rsmpi/rsmpi/actions/workflows/lint.yaml), the project currently seems to rely on the default settings.

I'm not sure what the best format spec would be, but I think there should be a `rustfmt.toml` file in the project so that the formatting is consistent for everyone.

Personally, I quite like the following formatter setup, and I think it produces very clean, readable code:

```toml
use_small_heuristics = "Max"
max_width = 80
wrap_comments = true

unstable_features = true
imports_granularity = "Crate"
group_imports = "StdExternalCrate"
```

- [`use_small_heuristiscs = "Max"`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#use_small_heuristics): Limits long lines in certain contexts. The actual length allowed differs for different constructs, and the defaults are pretty good.
- [`max_width = 80`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#max_width): The maximum line width.
- [`wrap_comments = true`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#wrap_comments): Whether to wrap comments. Although the default is off, having consistent line widths throughout the codebase can't hurt.
- [`unstable_features = true`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#unstable_features): Enables unstable features. This is probably a more controversial option, but it allows the two options below, which greatly improve import statements. This option and below may not be necessary, but having consistent, well-structured imports can be helpful. What do you guys think?
- [`imports_granularity = "Crate"`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#imports_granularity): This ensures that all imports are grouped by module, so you don't end up with any stragglers separated from other imports.
- [`group_imports = "StdExternalCrate"`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#group_imports): Separates out standard library modules, external modules and local/crate modules to more clearly show what comes from where. 

I understand that reformatting the codebase can cause issues, so maybe the best way forward is to construct a `rustfmt.toml` file that reflects the default settings currently used. That said, I think there's an argument to be made for keeping the codebase up-to-date with the latest Rust "standards". Additionally, any PRs affected by such a change should be able to copy the format file and run `cargo fmt` on their fork of the code (I think??).

Let me know what you think. I am happy to create the necessary PR if this change is welcome. Otherwise, it's easy enough to disable autoformatting in my editor :)

Pencilcaseman:
For more information regarding the format options selected, see issue #194 :)

### DIFF
--- a/build-probe-mpi/src/lib.rs
+++ b/build-probe-mpi/src/lib.rs
@@ -25,9 +25,9 @@
 
 mod os;
 
-pub use os::probe;
-
 use std::path::PathBuf;
+
+pub use os::probe;
 
 /// Result of a successfull probe
 #[allow(clippy::manual_non_exhaustive)]
--- a/build-probe-mpi/src/os/mod.rs
+++ b/build-probe-mpi/src/os/mod.rs
@@ -4,8 +4,7 @@ mod windows;
 #[cfg(unix)]
 mod unix;
 
-#[cfg(windows)]
-pub use self::windows::probe;
-
 #[cfg(unix)]
 pub use self::unix::probe;
+#[cfg(windows)]
+pub use self::windows::probe;
--- a/build-probe-mpi/src/os/unix.rs
+++ b/build-probe-mpi/src/os/unix.rs
@@ -9,9 +9,9 @@
 use core::fmt;
 use std::{env, error::Error, path::PathBuf, process::Command};
 
-use super::super::Library;
-
 use pkg_config::Config;
+
+use super::super::Library;
 
 #[derive(Debug, PartialEq)]
 struct UnquoteError {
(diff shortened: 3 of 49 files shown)


jedbrown (approved):
Thanks!

PR #202, opened 2025-01-30
feature: add feature 'libffi-system'

namachan10777:
It can be difficult to build `libffi` from source code on some HPC clusters. `libffi` has a feature that uses the system's libraries, which can be specified as a rsmpi feature.

I am working with a GH200 cluster (Miyabi of JCAHPC). The default compiler of this system is `nvc`, but it is difficult to compile `libffi` source code.
The ideal solution is to remove 'libffi' dependency from 'rsmpi or fix `libffi-rs`. However, maintenance of libffi-rs appears to be stalled and it is difficult to remove the libffi dependency from rsmpi.

I think this feature is also useful for other people who use rsmpi in environments that require special compilers.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,24 +21,24 @@ jobs:
           - os: ubuntu-latest
             rust: nightly
             mpi_package: libmpich-dev
-            cargo_flags: --all-features
+            cargo_flags: --features user-operations,derive,complex
             cargo_update: true
           - os: ubuntu-latest
             rust: stable
             mpi_package: libmpich-dev
-            cargo_flags: --all-features
+            cargo_flags: --features user-operations,derive,complex
           - os: ubuntu-latest
             rust: 1.70.0
             mpi_package: libmpich-dev
-            cargo_flags: --all-features
+            cargo_flags: --features user-operations,derive,complex
           - os: macos-latest
             rust: stable
             mpi_package: open-mpi
-            cargo_flags: --all-features
+            cargo_flags: --features user-operations,derive,complex
           - os: ubuntu-latest
             rust: stable
             mpi_package: libopenmpi-dev
-            cargo_flags: --all-features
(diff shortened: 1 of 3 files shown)


jedbrown (commented):
Thanks for submitting this. Is it not possible to use `--features=libffi/system` as described in the [features chapter](https://doc.rust-lang.org/cargo/reference/features.html#resolver-version-2-command-line-flags) or otherwise patch in that feature? If we make it a new feature of `mpi`, it'll need to be documented.

namachan10777:
Thank you for your comment.
I changed `README.md` to mention `libffi-system` feature in my patch.

My understanding is that the first solution it not applicable. `cargo` allows us to specify features for only directly dependent crates. I am writing a HPC application with `rsmpi` as a dependency, so `libffi` is a grandchild and I cannot specify a feature. I have now forked `rsmpi` and applied this my suggested patch.

> Is it not possible to use --features=libffi/system as described in the [features chapter](https://doc.rust-lang.org/cargo/reference/features.html#resolver-version-2-command-line-flags) or otherwise patch in that feature?

It is difficult to patch to `libffi` because `libffi` is overdue for maintenance.

However, as `libffi` seems to be used only for some of the collective functions, it may be more appropriate to close this PR so that it can be fully equivalent in a way that does not rely on `libffi`.

jedbrown:
Ah, fair point. What about writing this (in the top-level application) and letting feature unification apply it to what's used by rsmpi?
```toml
[dependencies]
libffi = { version = "3.2.0", features = ["system"] }
```
This is the recommended way to select a BLAS provider, for example, and avoids plumbing features through every library crate that transitively depends on BLAS: https://github.com/blas-lapack-rs/blas-lapack-rs.github.io/wiki (see bottom of page)

I'd also be open to removing dependence on libffi.

namachan10777:
I was able to solve the problem in my own way. I have realised that my proposal is unnecessary, so I will close the PR. Thank you very much.

PR #207, opened 2025-04-12
Update install directions in README

SirAlienTheGreat:
Includes specific instructions for fedora

I hope this is how you're supposed to contribute because I've never contributed to a public repo before


README.md
```diff
@@ -38,6 +38,7 @@ For a reasonable chance of success with `rsmpi` with any MPI implementation, you
   - `rsmpi` automatically uses `CRAY_MPICH_DIR` on Cray environments so the above need not be set
   - `mpich` and `ompi` are tried by default as a last resort
   - Tip: test with a command like `pkg-config --cflags --libs mpich`
+  - Tip: On Fedora 41, full installation with MPICH can be done by setting `$PKG_CONFIG_PATH` to `/usr/lib64/mpich/lib/pkgconfig` and `$LD_LIBRARY_PATH` to `/usr/lib64/mpich/lib/` after installing `mpich` through `dnf`
```
jedbrown, line 41: One can use `MPI_PKG_CONFIG` to tell rsmpi that you want to use it as first priority (this is documented in the README). I think manually setting `LD_LIBRARY_PATH` is only needed if you want to run the executables directly (instead of via `cargo mpirun` or `mpirun`). Can you check those and see if we can revise this PR to be more clear to an audience beyond the specific Fedora version?

README.md
```diff
@@ -38,6 +38,9 @@ For a reasonable chance of success with `rsmpi` with any MPI implementation, you
   - `rsmpi` automatically uses `CRAY_MPICH_DIR` on Cray environments so the above need not be set
   - `mpich` and `ompi` are tried by default as a last resort
   - Tip: test with a command like `pkg-config --cflags --libs mpich`
+  - The 2 easiest options for running rsmpi are through `cargo mpirun` or directly through `cargo run` (or the equivalent binary generated by `cargo build`)
+    - To use `cargo mpirun`, install it with `cargo install cargo-mpirun`, and then load mpi modules with `module load mpi/openmpi-x86_64`
```
jedbrown, line 42: You are writing into a bullet point that is underneath "export `MPI_PKG_CONFIG` to be ...". The suggestion to install `cargo-mpirun` can be moved out of this list and handled later (it's separate from installing rsmpi), and `cargo mpirun` usage can move into the `Usage` section.

README.md
```diff
@@ -38,6 +38,9 @@ For a reasonable chance of success with `rsmpi` with any MPI implementation, you
...
+    - To run the binary directly, you need to set 2 environmental variables to paths that may vary between distros. On Fedora 41, this involves setting `$PKG_CONFIG_PATH` to `/usr/lib64/mpich/lib/pkgconfig` and `$LD_LIBRARY_PATH` to `/usr/lib64/mpich/lib/` after installing `mpich` through `dnf`
```
jedbrown, line 43: `PKG_CONFIG_PATH` has no effect after building (and one is supposed to use `MPI_PKG_CONFIG`, as in the parent that this is organized under) because it ensures that this path will be used to find MPI, not only as a fallback option).

I think the Usage section should explain that one can run with `cargo mpirun -n 2 --bin myapp` or `mpiexec -n 2 target/debug/myapp` with no further changes. But if one wants to just run `target/debug/myapp`, then you'll need to set `LD_LIBRARY_PATH`. On MacOS, that path will presumably be `DYLD_LIBRARY_PATH`. I don't know what is necessary on Windows.

Note that the same, in general, is necessary for C code, but the `mpicc` wrapper will (for common implementations), set RPATH (see `mpicc -show` output). It's not possible to set RPATH for a library crate, but one can set it for a binary crate [see this example](https://gitlab.com/petsc/petsc-rs/-/blob/main/examples/snes/build.rs). More discussion here (for a related package):

https://gitlab.com/petsc/petsc-rs#finding-libpetsc-at-run-time

SirAlienTheGreat: How do you get `MPI_PKG_CONFIG` to work without `PKG_CONFIG_PATH`? It seems that `build_probe_mpi::probe()` calls the `pkg-config` library, which relies on having `PKG_CONFIG_PATH` set, or errors with something like:

```
Could not find MPI library for various reasons:

  Reason #0:

  pkg-config exited with status code 1
  > PKG_CONFIG_PATH= PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 pkg-config --libs --cflags mpich

  The system library `mpich` required by crate `mpi-sys` was not found.
  The file `mpich.pc` needs to be installed and the PKG_CONFIG_PATH environment variable must contain its parent directory.
  PKG_CONFIG_PATH contains the following:
      - 

  HINT: you may need to install a package such as mpich, mpich-dev or mpich-devel.
```

jedbrown: https://github.com/rsmpi/rsmpi/blob/main/build-probe-mpi/src/os/unix.rs#L116-L117

For example:
```console
$ pkg-config --libs /opt/mpich/lib/pkgconfig/mpich.pc
-Wl,-rpath -Wl,/opt/mpich/lib -Wl,--enable-new-dtags -L/opt/mpich/lib -lmpi -Wl,-O1 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,-z,pack-relative-relocs -flto=auto -lpthread -lhwloc -lucp -lucs
```
How did you set `MPI_PKG_CONFIG`?

PR #209, opened 2025-04-21
Update libffi to 4.0.0

0pendansor:
LIbffi is broken in MacOS 15+. The clang compiler that comes with Xcode that comes with MacOS 15 is incompatible with libffi 3.2.0. 

There has been a fix, and it is updated in libffi 4.0.0. See https://github.com/tov/libffi-rs/issues/109

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ complex = ["dep:num-complex"]
 
 [dependencies]
 conv = "0.3.3"
-libffi = { version = "3.2.0", optional = true }
+libffi = { version = "4.0.0", optional = true }
 memoffset = { version = "0.9", optional = true }
 mpi-derive = { path = "mpi-derive", version = "0.1.2", optional = true }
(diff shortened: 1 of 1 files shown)


jedbrown (commented):
Thanks, I started the pipeline. Is it worth changing CI (.github/workflows/test.yaml) so that this would be tested?

jedbrown:
I fixed CI issues. Can you please rebase on main?

jedbrown:
This bumps MSRV to 1.78.

Cargo.toml
```diff
@@ -37,7 +37,7 @@ complex = ["dep:num-complex"]
 
 [dependencies]
 conv = "0.3.3"
-libffi = { version = "3.2.0", optional = true }
+libffi = { version = "4.0.0", optional = true }
```
jedbrown, line 40:
```suggestion
libffi = { version = "4.1.0", optional = true }
```
I see a new version is out. And please update the `rust-version` field; then we should be able to merge this.

PR #203, opened 2025-02-11
Add bindgen-runtime feature

mscroggs:
This will allow this bindgen default feature `runtime` to be disabled by using `default-features=false`. This would resolve the issue https://github.com/LIHPC-Computational-Geometry/scotch-rs/issues/12 in scotch-rs

PR #214, opened 2025-06-06
deps: libffi-4.1 and others

jedbrown:
The updated libffi is needed to fix an issue with MacOS 15, and the fix has a side-effect of increasing MSRV to 1.78. See
https://github.com/rsmpi/rsmpi/pull/209


.github/workflows/test.yaml
```diff
@@ -55,6 +55,8 @@ jobs:
         run: |
           brew install llvm@19 ${{ matrix.mpi_package }}
           echo TMPDIR=/tmp | tee -a $GITHUB_ENV
+          echo PATH="/opt/homebrew/opt/llvm@19/bin:$PATH" | tee -a $GITHUB_ENV
+          echo DYLD_LIBRARY_PATH=$(/opt/homebrew/opt/llvm@19/bin/llvm-config --libdir) | tee -a $GITHUB_ENV
```
jedbrown, line 59: @mscroggs This PR is now a working alternate to #203. It works out of the box on Linux and Windows (`runtime` feature disabled), but on MacOS, it requires setting `DYLD_LIBRARY_PATH` whenever the `runtime` feature is disabled. Can you comment about whether this meets your needs, and whether you have a different recommendation to make it work without needing this environment variable?

tbetcke: Hi, just ran into this myself and found this PR. What is the current status? I noticed there is also another PR #216 on this topic. How do they relate to each other?

tbetcke: The DYLD_LIBRARY_PATH issue will confuse users. It would be good to have a default configuration on Mac that does not require it. But I am not sure why it is needed at all and where it is coming from.

PR #224, opened 2026-06-14
make is_initialized and is_finalized public

philbucher:
Hi,
I propose to change the visibility of `is_initialized` and `is_finalized`
I use this to check the status of MPI before interacting with an external library that requires MPI to be initialized
Thanks!


jedbrown:
Thanks, can you add the `# Standard section(s)` like all the other public functions?

Note that when solely using rsmpi, one can just call `initialize()` or whatever because those return an `Option<Universe>` (in case MPI was already initialized). And that if one has a multithreaded application in which MPI is called independently from rsmpi, there is a race condition between checking `is_initialized()` and initializing. (I don't think this is a soundness issue so no need for `unsafe`, but it would lead to `MPI_Initi_thread()` failing, so it isn't useful.) If you could write a note about that, it would help prevent surprises with the sort of code patterns one might be drawn to when using these rather low-level/stateful interfaces.

philbucher:
Thanks for the review! I checked the standard and improved the documentation.

philbucher:
hi @jedbrown , did you already have the chance to review the changes?
Thanks a lot for your time

jedbrown:
Thank you! Sorry about the delay.

philbucher:
no worries and thanks for taking the time!

